PR #2, opened 2022-06-18
.github/workflows: Automatically build package

mrnuke:
Add a hacky github workflow file to automatically build an ipk on
pushes or pull requests. This is barely held together by duct tape and
nails. It's not supposed to be pretty.

Under the skirt, this works by running the openwrt SDK under a docker
container. The realtek-poe package is dropped under the "packages/"
dir of the SDK, enabling the package to build as if it were a core
package. This is a simple, albeit unelegant sulotion to get an ipk.


Hurricos:
Looks good. Merged.

PR #3, opened 2022-06-19
realtek-poe: Fix status for devices with more than 8 PoE ports

mrnuke:
Only status data for the first 8 ports was requested. Most devices
have 8 ports, and the reply packet has room for 8 ports. It was the
perfect match!

People quickly figured out realtek-poe works for unicorn switches with
more than 8 PoE ports. And so they did. And they increased MAX_PORT
without regards to the out-of-bounds access problem they created. And
realtek-poe refused to crash!

Stop reading past the end of the packet in poe_reply_port_overview().
Just requests more packets if there are more than 8 ports.

Fixes: https://github.com/Hurricos/realtek-poe/commit/8c429f1d17e1e5b063043ee06a791fcb2bfcaf20 ("realtek-poe: Increment up to MAX_PORT, not 8")

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -467,18 +467,18 @@ poe_reply_port_ext_config(unsigned char *reply)
 	return 0;
 }
 
-/* 0x2a - Get all port status */
-static int
-poe_cmd_port_overview(void)
+/* 0x28 - Get all all port status */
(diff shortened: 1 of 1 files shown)


Hurricos:
As far as I can tell, this PR is superseded by #6, so I'm closing it.

Re-open if I'm wrong. :^)

mrnuke:
You are correct. I had expected github to detect this and auto-close this PR.

PR #4, opened 2022-06-21
Cleanup code madness, round 1

mrnuke:
Main point of this PR is to eliminate ~50 lines of code while maintaining functionality.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -37,8 +37,8 @@ struct port_config {
 struct config {
 	int debug;
 
-	int budget;
-	int budget_guard;
+	float budget;
+	float budget_guard;
 
 	int port_count;
 	struct port_config ports[MAX_PORT];
@@ -391,7 +391,7 @@ poe_cmd_power_mgmt_mode(unsigned char mode)
 
 /* 0x18 - Set global power budget */
 static int
-poe_cmd_global_power_budget(int budget, int guard)
+poe_cmd_global_power_budget(float budget, float guard)
 {
 	unsigned char cmd[] = { 0x18, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
 
--- a/src/main.c
+++ b/src/main.c
@@ -66,8 +66,6 @@ struct cmd {
 	unsigned char cmd[12];
 };
 
-static struct uloop_timeout state_timeout;
-static struct ubus_auto_conn conn;
 static struct ustream_fd stream;
 static LIST_HEAD(cmd_pending);
 static unsigned char cmd_seq;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -391,7 +391,7 @@ poe_cmd_power_mgmt_mode(unsigned char mode)
 
 /* 0x18 - Set global power budget */
 static int
-poe_cmd_global_power_budget(int budget, int guard)
+poe_cmd_global_power_budget(float budget, float guard)
```
Hurricos, line 345: `write16_be` [expects](https://github.com/Hurricos/realtek-poe/blob/db485c85b2c6c997049b81c44a3c0d6d57d1728c/src/main.c#L400) a `uint16_t`, yet this change causes it to now be passed a float, twice ([`float * int = float`](https://stackoverflow.com/a/17057689)).

I'd ask you to cast, but I'm more alarmed by the fact that this code compiles and works.

src/main.c
```diff
@@ -89,99 +89,50 @@ static void write16_be(uint8_t *raw, uint16_t value)
 	raw[1] =  value & 0xff;
 }
 
-static void
```
Hurricos, line 94: I think I've seen this pattern elsewhere in OpenWrt uci parsing code -- maybe it exists to standardize it and make it machine-readable? -- but --

1) I might be wrong and this part of `realtek-poe` might just be a blogicism
2) Even if I found this pattern in other uci / ubus parsing code, I don't think I'd be able to call up the right people to explain their patterns.

So this has my blessing.

Hurricos (commented):
Added just a bit of review. That which isn't over my head looks good -- I will say that my tweaks to the code to enable sending frames through ubus were very much mechanically added, I believe by adapting ubus hooks from wpa_supplicant.

Hurricos:
Feedback given. Tweak to cast those floats before `write16_be` (or tell me what I'm missing), and we can merge this and take a look at merging #6 using fewer commits.

src/main.c
mrnuke: There's an implicit int conversion. That is valid C.

Now there is an issue if the result is outside of [0, UINT16_MAX]. But that issue also existed before this patch,

mrnuke: Here's the wording from the C standard, for those playing along at home:

```
6.3.1.4 Real floating and integer
1 When a finite value of real floating type is converted to an integer type other than _Bool,
the fractional part is discarded (i.e., the v alue is truncated to ward zero). If the v alue of
the integral part cannot be represented by the integer type, the behavior is undefined.
```

Hurricos: *That's* what I'm missing :^)

Hurricos:
I am comfortable with this; the lack of WIP suggests you are as well. Merged.

PR #8, opened 2022-06-28
 realtek-poe: Get port count from config, not system status packet

mrnuke:
The initial assumption that there are eight PoE ports is not always
valid. This assumption was corrected with the receipt of the first
status packet, but that is too late. When this packet arrives, the
initial config will have already happened. Thys ports above 8 do not
get configured. One might dare call this a race condition.

Use the config file to determine the port count. Remove the assumption
that the system has eight ports. It is not problematic if the config
has more ports than the sytem. The MCU will respond with all f's to
commands for invalid ports.

The case where the system has more ports than the config is not
handled. It doesn't have to be. The "poe info" command only sends data
for configred ports that are enabled. There's no point in probing
anything more.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -40,7 +40,7 @@ struct config {
 	float budget;
 	float budget_guard;
 
-	int port_count;
+	unsigned int port_count;
 	uint8_t pse_id_set_budget_mask;
 	struct port_config ports[MAX_PORT];
 };
@@ -674,7 +674,7 @@ poe_stream_open(char *dev, struct ustream_fd *s, speed_t speed)
 static int
 poe_port_setup(void)
 {
-	int i;
+	size_t i;
 
 	for (i = 0; i < config.port_count; i++) {
 		if (!config.ports[i].enable) {
@@ -729,7 +729,7 @@ poe_initial_setup(void)
 static void
 state_timeout_cb(struct uloop_timeout *t)
 {
-	int i;
+	size_t i;
 
 	poe_cmd_power_stats();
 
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -112,6 +113,7 @@ static void load_port_config(struct uci_context *uci, struct uci_section *s)
 		ULOG_ERR("invalid port id=%lu for %s", id, name);
 		return;
 	}
+	config.port_count = MAX(config.port_count, id);
```
Hurricos, line 116: I was a little worried about a race condition against a callback after a reload if the port count changes, but I don't think there's reason to worry about it here.

Hurricos:
Works for me on my ZyXEL GS1900-24HPv1.

Tested-by: Martin Kennedy <hurricos@gmail.com> (not gonna bother adding that to the commits, though).

src/main.c
mrnuke: This may be asynchronous code, but it is sequential. You're guaranteed to exit `load_port_config()` before you reach other parts of the code that use this config. You don't need to be re-entrant or thread-safe because you're not being pre-empted or context-swapped by some other task.

Hurricos: I was thinking something like `ubus call poe info` ... but now I'm realizing how little I know/knew about how ubus is written -- I'm guessing that the consumption of tasks is performed by the daemon exposing the callback through its use of a `uloop`? Rather than there being some kind of shared object code being exposed to the `ubus` binary here.

mrnuke: There's one event loop. There's only one function ever getting called at any given time. If you look carefully, you won't see any blocking function. No sleep() or wait_for(). If a function were to block, then we couldn't process events for that amoutn of time. ubus events are no different. In the end you're waiting fir events on a file descriptor or socket.

PR #14, opened 2022-07-03
realtek-poe: Do not disable PoE on all ports during setup

Hurricos:
realtek-poe: Do not disable PoE on all ports during setup

When `/etc/init.d/poe restart` is invoked, the startup process calls
`poe_cmd_global_port_enable(0)`, which causes all PoE sinks connected
to the powering device will temporarily lose power and reboot.

Fix this by removing that `poe_cmd_global_port_enable(0)` from
`poe_initial_setup`.

Also, since this is the only use of `poe_cmd_global_port_enable`,
remove it entirely.

Signed-off-by: Martin Kennedy <hurricos@gmail.com>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -285,20 +285,6 @@ static int poe_cmd_port_mapping_enable(bool enable)
 	return poe_cmd_queue(cmd, sizeof(cmd));
 }
 
-/* 0x06 - Set global port enable
- *	0: Disable PSE Functionality on all Ports
- *	1: Enable PSE Functionality on all Ports
- *	2: Enable Force power Functionality on all ports
- *	3: Enable Force Power with Disconnect Functionality on all Ports
- */
-static int
-poe_cmd_global_port_enable(unsigned char enable)
-{
-	unsigned char cmd[] = { 0x06, 0x00, enable };
-
-	return poe_cmd_queue(cmd, sizeof(cmd));
-}
-
 /* 0x10 - Set port detection type
  *	1: Legacy Capacitive Detection only
  *	2: IEEE 802.3af 4-Point Detection only (Default)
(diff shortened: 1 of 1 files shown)


mrnuke:
Are you sure we want to add `poweron` and `poweroff` at this point?

Picturing this in a luci intreface, such as "Network->Switch", Do we want a big button "Turn off all ports"? Though I see the usefulness of having a way to cut off power, that's also easy to type.

I propose one of the following two things:

1. Remove `poe_cmd_global_port_enable()` for now. It's trivial to add later.
2. squash the two commits into one, and feel free to take credit. You have a point between the first and second that the code won't compile. I'd rather avoid that.

Hurricos:
> 2\. You have a point between the first and second that the code won't compile. I'd rather avoid that.

You might be mistaking the order. 5822e926d16d33449eb1a66148484cbf15b50b6d is the parent of d3ab2e9eaae70e7f04d8bd471ce3120b64d4351c. I did this intentionally to avoid a non-compiling commit.

RE: 1):
- It's annoying to remove and add code to satisfy the compiler gods,
- We ought to expose whatever functionality we don't use to `ubus`,

... but:

- I can see that the commit text in 5822e926d16d33449eb1a66148484cbf15b50b6d is more drafty than you'd like.
- Leaving half-baked functionality is bad; exposing functionality should be done intentionally as part of a larger patchset to change the model of interacting with the MCU.
- Reducing LOC is good

Hurricos:
Force-pushed everything back, dropping 5822e926d16d33449eb1a66148484cbf15b50b6d from the patchset.

mrnuke:
My previous argument was not about the theological principle of appeasing the compiler gods. It has more to do with the practical aspects of maintaining causality for `git bisect` operations.

If you're willing to drop the URLs and parenthetical references (with one or more parentheses) from the commit message, that would be awesome. Otherwise :+1:

Hurricos:
> If you're willing to drop the URLs and parenthetical references (with one or more parentheses) from the commit message, that would be awesome. Otherwise +1

Advice taken.

Issue #13, opened 2022-07-03
realtek-poe almost always fails to configure final ports due to busy system at early boot


Hurricos:
## Issue identification

On reboot, I pretty much always see:

```txt
root@OpenWrt:~# ubus call poe info | grep Disabled -A 1 -B 3
		"lan15": {
			"priority": 2,
			"mode": "PoE",
			"status": "Disabled"
		},
--
		"lan18": {
			"priority": 2,
			"mode": "PoE",
			"status": "Disabled"
		},
		"lan19": {
			"priority": 2,
			"mode": "PoE",
			"status": "Disabled"
		},
		"lan20": {
			"priority": 2,
			"mode": "PoE",
			"status": "Disabled"
		},
		"lan21": {
			"priority": 2,
			"mode": "PoE",
			"status": "Disabled"
		},
		"lan22": {
			"priority": 2,
			"mode": "PoE",
			"status": "Disabled"
		},
		"lan24": {
			"priority": 2,
			"mode": "PoE",
			"status": "Disabled"
		}
```

Hurricos:
## Debugging
~~I suspect this is because attempts to write out to `/dev/ttyS1` are failing because a lot is going on at this point in the boot process.~~

Ultimately: it seems this is because the MCU just isn't ready at the given point in boot (see [below](https://github.com/Hurricos/realtek-poe/issues/13#issuecomment-1173165793)).

## Solutions
- [ ] Increase `files/etc/init.d/poe::START` from `80` to `99`
  - This seems to work (5 tests of each confirm this) but does nothing for the underlying issue.
- [ ] A mechanism to re-queue failed commands.
  - So far, that's the most promising lead.

mrnuke:
I can repro the issue without needing to reboot. It appears problem is related to sending too many packets too fast to the MCU. Eventually it replies with "not ready".

I've been experimenting with a mechanism to re-queue failed commands. So far, that's the most promising lead.

mrnuke:
No PR exists yet for this code, but to demo the issue:

```
root@engenius-2910:~# /tmp/realtek-poe 
realtek-poe: MCU didn't like our command because incomplete request
realtek-poe:    CMD:    13 2c 05 02 ff ff ff ff ff ff ff 3f
realtek-poe:    reply:  fd ff ff ff ff ff ff ff ff ff ff f3
realtek-poe: Ate 1 replies
realtek-poe: Ate 0 replies
realtek-poe: Chomp/chomp/chomp
[...]

Issue #10, opened 2022-06-28
`ubus call poe info` status is randomly stuck on (cold/re)boot to 'unknown' for all ports


walterav1984:
ZyXEL GS1900-24HPv1 
realtek-poe CI22 version

Sometimes it shows correct realtime status info after (cold/re) boot for the `ubus call poe info` and [sometimes](https://forum.openwrt.org/t/bringing-support-for-realtek-poe-to-mainline-openwrt/127777/106) its stuck in `unknown` state for all ports although they are actually `Delivering power` fine with PoE devices powered on. A `/etc/init.d/poe restart` is the only remedy to make the status correct again with the downside of temporary disabling PoE.

mrnuke:
I'm going to abstain from speculating until we can get some debug logs (`realtek-poe -d`).

walterav1984:
Trying to redirect the output from `realtek-poe -d > output.log` failed. Therefor I tried `realtek-poe -d 2>&1 | grep X > realtek-poe-CI22-p2-24even.log` and it logs the output. 

The laptop was connected to port 3 all the time, 1 ssh session for log the 2nd ssh session for `ubus call poe info`. Than a single PoE device was connected and disconnected until the PoE comes on, for every even port starting at port 2 and finished on 24. After port 14 the `ubus call poe info` status hang as in it reported that port 14 was delevering power and port 16-24 kept `Searching` ... but ports 16-24 did power on the PoE although their status remaind `Searching`. After disconnecting the last PoE device at port 24 CTRL-C the deamon. Did the same test for uneven ports 1-23(laptop on port2), funny thing is this time `ubus call poe info` hangs correct reporting after port5 but like before port 7-23 power on PoE fine but status keeps `Searching` while port 5 stayed on `Delevering Power`. 

even p2-p24
https://pastebin.com/j0TSm7Hx

uneven p1-p23 part1 / part2 pastebin upload size limit 
https://pastebin.com/wfWiht19
part1 11676~
https://pastebin.com/xyz2Tkjg
part2 ~16526

mrnuke:
### Analysys

There's your problem:

```
TX -> 26 17 11 ff ff ff ff ff ff ff ff 46
RX <- fe ff ff ff ff ff ff ff ff ff ff f4
TX -> 30 18 11 ff ff ff ff ff ff ff ff 51
RX <- fe ff ff ff ff ff ff ff ff ff ff f4
```

[...]

Issue #9, opened 2022-06-28
random re(connecting) PoE device may put some ports in sticky Disabled state which won't power on


walterav1984:
ZyXEL GS1900-24HPv1 
realtek-poe CI22 version

More specific info will come later but both [@Hurricos](https://github.com/Hurricos/realtek-poe/pull/6#issuecomment-1166656308) "sees certain ports as 'disabled'" and [myself](https://github.com/Hurricos/realtek-poe/pull/6#issuecomment-1167638946) already encountered this behaviour and its random and will not always manifest.

Although the `/etc/config/poe` file may have these ports set to Enabled, the `ubus call poe info` lists them as Disabled and they won't poweron a PoE device anymore although they did PoE before the random state change (after manually re/dis connecting PoE device). Doing a reboot in the past won't solve this either, also reloading service or ubus won't help. Only doing a `/etc/init.d/poe restart` may wake them up again.

I reported [earlier](https://github.com/Hurricos/realtek-poe/pull/6#issuecomment-1167638946) see the end of that comment that with previous CI14 I also needed to change the `/etc/config/poe` file from enabled > disabled > enabled before port would poweron.

mrnuke:
When this happens, please try the following:

    # logread -te realtek-poe

walterav1984:
Repost of commit on issue #10 

This time I do however have a log now of the funny sticky disabled ports 4 & 7 (possible other github issue #9 ), after reboot. So this time the output is after a reboot not coldboot. PoE power comes back on after reboot although no realtek-poe service is started so no ubus output but after starting realtek-poe -d this was the output.

```
ubus call poe info #after reboot PoE power come on automatically without realtek-poe service running 
#but as soon as realtek-poe -d started to get to ubus it resets all PoE power and this was the output
{
	"firmware": "v17.1",
	"mcu": "ST Micro ST32F100 Microcontroller",
	"budget": 170.000000,
	"consumption": 0.000000,
	"ports": {
		"lan1": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Searching"
		},
		"lan2": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Searching"
		},
		"lan3": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Searching"
		},
		"lan4": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Disabled"   <--- Not disabled but PoE device won't power on again on this port!
		},
		"lan5": {
[...]

PR #15, opened 2022-07-04
 realtek-poe: Implement a retry mechanism for failed commands

mrnuke:
The MCU doesn't like receiving a flood of commands. If it receives a
large number of commands in rapid succession, it may refuse to process
them. The problem is intermittent, but can be identified by receiving:

     fd ff ff ff ff ff ff ff ff ff ff f3

Testing shows that the MCU sends this packet several times, even if it
receives no further commands. There is no "I am ready" packet, which
makes this shituation more cumbersome to get out of.

A simple mechanism is to wait a certain amount of time for the MCU to
recover, and then start sending commands again. Keep the current
command on the queue, and only resend it after this timeout expires.
The hope is that this gives the MCU enough time to recuperate.

The current implementation recycles the sequence number when sending
retries. That could be problematic in matching the reply and request.
However, in practice this has not shown to be a problem.

Implement a 5 retry policy with a 100ms timeout for each failure. This
waits about half a second before giving up on the current command and
sending the next command in the queue.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -26,6 +26,7 @@ typedef int (*poe_reply_handler)(unsigned char *reply);
 /* Careful with this; Only works for set_detection/disconnect_type commands. */
 #define PORT_ID_ALL	0x7f
 #define MAX_PORT	24
+#define MAX_RETRIES	5
 #define GET_STR(a, b)	((a) < ARRAY_SIZE(b) ? (b)[a] : NULL)
 #define MAX(a, b)	(((a) > (b)) ? (a) : (b))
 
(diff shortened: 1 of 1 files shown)


mrnuke:
If this **cough** fix **cough** works, I'm expecting this to address #9 and #13, as well as #10 for the most part. I've also noticed a de-synchronized packet issue in #10, which is not addressed here.

Hurricos:
OK, I like this! After testing, I noticed that I was overall only getting a small handful of MCU error messages, and I *had* to go read the source to find out.

I really think the 100ms pause is a great idea, if the MCU is going to complain we shouldn't be hammering on it. I like how this is implemented.

Rebasing and merging!

PR #17, opened 2022-07-06
realtek-poe: Do not abandon status for disabled ports

### LINKED ISSUE #16: disabling a PoE port in `/etc/config/poe` removes the whole port status from `ubus call poe info`
ZyXEL GS1900-24HPv1
realtek-poe all versions affected till CI31 

Bug or feature?

This has been around from the beginning of testing the realtek-poe package. Although we have seen ports being flagged unwanted as [status Disabled](https://github.com/Hurricos/realtek-poe/issues/9) in the past and those ports were not delivering PoE anymore. 

Currently Disabling and Enabling ports via `/etc/config/poe` does work but shouldn't the `ubus call poe info` reflect the port status as Disabled instead of not showing the port at all when disabled?

```
$ cat /etc/config/poe #snippet
config port
        option enable   '1'
        option id       '23'
        option name     'lan23'
        option poe_plus '1'
        option priority '2'

config port
        option enable   '1'
        option id       '24'
        option name     'lan24'
        option poe_plus '1'
        option priority '2'

$ ubus call poe info #snippet
		},
		"lan23": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Searching"
		},
		"lan24": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Delivering power",
			"consumption": 2.000000
		}

$ cat /etc/config/poe #snippet
config port
        option enable   '1'
        option id       '23'
        option name     'lan23'
        option poe_plus '1'
        option priority '2'

config port
        option enable   '0'
        option id       '24'
        option name     'lan24'
        option poe_plus '1'
        option priority '2'

$ ubus call poe info #snippet
		},
		"lan23": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Searching"
		}
> The disabled PoE port 24 is just missing
```

Shouldn't it look like this when disabling a PoE port?
```
$ ubus call poe info #snippet
		},
		"lan23": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Searching"
		},
		"lan24": {
			"priority": 2,
			"mode": "PoE+",
			"status": "Disabled"
		}


```

mrnuke:
A port may be disabled, but it is still part of the family. It exists,
it consumes space, and it might even carry network traffic.

`ubus call poe info` should reflect that. Instead, it ignores ports
configured as disabled. In many jurisdictions, this is considered port
abuse, and users might get a visit from port protection services.

To resolve this, introduce the concept of a `valid` port config, and
keep all valid ports in the "poe info" packet.

Use a bitfield for the "valid" and "enable" flags to make sure the
port_config struct does not increase in size.

Should resolve #16

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -29,7 +29,8 @@ typedef int (*poe_reply_handler)(unsigned char *reply);
 
 struct port_config {
 	char name[16];
-	unsigned char enable;
+	unsigned int valid : 1;
+	unsigned int enable : 1;
 	unsigned char priority;
 	unsigned char power_up_mode;
 	unsigned char power_budget;
@@ -117,6 +118,7 @@ static void load_port_config(struct uci_context *uci, struct uci_section *s)
 	id--;
 
 	strncpy(config.ports[id].name, name, sizeof(config.ports[id].name));
+	config.ports[id].valid = 1;
 	config.ports[id].enable = enable ? !strcmp(enable, "1") : 0;
 	config.ports[id].priority = priority ? strtoul(priority, NULL, 0) : 0;
 	if (config.ports[id].priority > 3)
@@ -754,7 +756,7 @@ ubus_poe_info_cb(struct ubus_context *ctx, struct ubus_object *obj,
 	for (i = 0; i < config.port_count; i++) {
 		void *p;
 
-		if (!config.ports[i].enable)
+		if (!config.ports[i].valid)
 			continue;
 
 		p = blobmsg_open_table(&b, config.ports[i].name);


Hurricos:
I like it! I haven't read this closely, but this should also solve the question of reading status off of ports not explicitly noted in the configuration, right?

If so, we could later re-enable command 0x06 and provide a less intensive configuration pathway.

mrnuke:
If it's not noted in the config, it shouldn't show up in the output.

Hurricos:
> If it's not noted in the config, it shouldn't show up in the output.

OK, got it. No problem; gathering info from all ports the MCU asserts to exist could fold in nicely to a later configuration streamlining effort.

mrnuke:
I presume you could `MAX(config.port_count, state.num_detected_ports)` to make sure you query all ports. I had considered that when writing #8, but you wouldn't get the output with `poe info` anyway. The other option, pushing status for all detected ports would break the config model.

I guess you could export `num_detected_port` via `poe info, but what is the use case?

walterav1984:
This seems fixed for CI36 but not CI40 and CI41...

mrnuke:
> This seems fixed for CI36 but not CI40 and CI41...

That is expected. CI 36 is the correct build for this PR. Each PR gets its own CI build. A higher CI number doesn't necessarily mean more better.

walterav1984:
Thanks, I was in the assumption that every CI build accumulated the previous patches.

Hurricos:
Tests OK on my hardware.

Simple one to merge. Done.

Issue #18, opened 2022-07-06
Some MCU commands are used inefficiently


mrnuke:
The following MCU commands take a list of ports:

- `10` - poe_cmd_port_detection_type()
- `11` - poe_cmd_port_classification()
- `13` - poe_cmd_port_disconnect_type()
- `15` - poe_cmd_port_power_limit_type()
- `16` - poe_cmd_port_power_budget()
- `1A` - poe_set_port_priority()
- `1C` - poe_set_port_power_up_mode()

realtek-poe is sending only one port per command. The issue is we're sending up to 192 config packets for a 24-switch port. That could drop to 66 by shoving more than one port per packet. I suspect this might help with #10 and #13.

### Further optimizations

- poe_cmd_port_detection_type() and poe_cmd_port_disconnect_type() can also use 0x7f to signal all ports.
- The trace that I have for 1900 ZyXEL GS1900-24HPv1 doesn't make use of `00`, `10`, `11`, or `13`.

mrnuke:
FIxed by #19

PR #19, opened 2022-07-09
realtek-poe: Reduce number of port setup packets

mrnuke:
In response to #18, try to be more efficient with port setup.

Use 4 port commands efficiently
------------------------------------------

Most commands used for configuring port parameters can set up tp four
ports per packet. Only setting one port port and setting all other
fields to 0xff is woefully inefficient. How ineffiecient? On a 24-port
switch, each command could be sent in 6 packets instead of 24.

To achieve this, pack 4 port IDs and values in each command, and skip
unused or disabled ports.

I had considered that unpacking an array of structures, to pack into
two arrays to then repack into one array is "woefully inefficient".
The CPU and RAM is 5 orders of magnitude faster than the serial link,
and I was going for the least unreadable approach. I'm open to
improvements as long as it reduces the line count.

Use port ID 0x7f (all ports) where supported
----------------------------------------------------------

The set_port_disconnect_type and set_port_detection_type commands can
configure all ports in one go if passed a port_id of 0x7f. On 24-port
switches, using this approach significantly reduces the number of
packets sent during configuration.

There is no anticipated need to change the config on a per-port basis.
Only send one of each poe_cmd_port_disconnect_type() and
poe_cmd_port_detection_type() commands, with a port ID of 0x7f.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -23,6 +23,8 @@
 
 typedef int (*poe_reply_handler)(unsigned char *reply);
 
+/* Careful with this; Only works for set_detection/disconnect_type commands. */
+#define PORT_ID_ALL	0x7f
 #define MAX_PORT	24
 #define GET_STR(a, b)	((a) < ARRAY_SIZE(b) ? (b)[a] : NULL)
 #define MAX(a, b)	(((a) > (b)) ? (a) : (b))
@@ -686,6 +688,9 @@ poe_port_setup(void)
 {
 	size_t i;
 
+	poe_cmd_port_disconnect_type(PORT_ID_ALL, 2);
+	poe_cmd_port_detection_type(PORT_ID_ALL, 3);
+
 	for (i = 0; i < config.port_count; i++) {
 		if (!config.ports[i].enable) {
 			poe_cmd_port_enable(i, 0);
@@ -700,9 +705,7 @@ poe_port_setup(void)
 		} else {
 			poe_cmd_port_power_limit_type(i, 1);
 		}
-		poe_cmd_port_disconnect_type(i, 2);
 		poe_cmd_port_classification(i, 1);
-		poe_cmd_port_detection_type(i, 3);
 		poe_cmd_port_enable(i, 1);
 	}
 
--- a/src/main.c
+++ b/src/main.c
@@ -262,6 +262,14 @@ poe_cmd_queue(unsigned char *_cmd, int len)
 	return 0;
 }
 
+static int poet_cmd_4_port(uint8_t cmd_id, uint8_t port[4], uint8_t data[4])
+{
+	uint8_t cmd[] = { cmd_id, 0x00, port[0], data[0], port[1], data[1],
+					port[2], data[2], port[3], data[3] };
+
+	return poe_cmd_queue(cmd, sizeof(cmd));
+}
+
 /* 0x00 - Set port enable
  *	0: Disable
  *	1: Enable
@@ -300,12 +308,9 @@ poe_cmd_port_detection_type(unsigned char port, unsigned char type)
  *	0: Disable
  *	1: Enable
  */
-static int
-poe_cmd_port_classification(unsigned char port, unsigned char classification)
+static int poe_cmd_port_classification(uint8_t port[4], uint8_t enable[4])
 {
-	unsigned char cmd[] = { 0x11, 0x00, port, classification };
-
-	return poe_cmd_queue(cmd, sizeof(cmd));
+	return poet_cmd_4_port(0x11, port, enable);
 }
 
 /* 0x13 - Set port disconnect type
@@ -328,12 +333,9 @@ poe_cmd_port_disconnect_type(unsigned char port, unsigned char type)
  *	1: Class based. The power limit for class 4 devices is determined by the high power limit.
  *	2: User defined
  */
-static int
-poe_cmd_port_power_limit_type(unsigned char port, unsigned char type)
+static int poe_cmd_port_power_limit_type(uint8_t port[4], uint8_t limit[4])
 {
-	unsigned char cmd[] = { 0x15, 0x00, port, type };
-
-	return poe_cmd_queue(cmd, sizeof(cmd));
+	return poet_cmd_4_port(0x15, port, limit);
 }
 
 /* 0x16 - Set port power budget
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -663,6 +659,43 @@ poe_stream_open(char *dev, struct ustream_fd *s, speed_t speed)
 	return 0;
 }
 
+static int poet_setup(const struct port_config *ports, size_t num_ports)
```
svanheule, line 682: Happy typo? :)

svanheule:
Instead of explicitly building 4-port commands, have you considered coalescing frames in `poe_cmd_queue()`?

The queueing function could check a list of command IDs that can be coalesced. If there is nothing queued, a frame can be sent immediately. Otherwise, you could:
1. check the last queued frame (or the entire list) for (frames with) the same ID,
2. check if the already queued frame has room available for pooling ports,
3. de-queue the old frame, update, and re-queue
4. queue the shortened new frame, or drop it if all ports were coalesced into the old one.

Just a thought, not sure if it would make the code shorter or easier to parse.

src/main.c
mrnuke: Once you find the [poem.py parser](https://github.com/Hurricos/realtek-poe/blob/mrnuke-devel/util/poem.py) you begin to realize that this is quite intentional. This is the poet that writes the serial data poem.

mrnuke:
@svanheule, I have not considered such approach until you brought it up.

#### Algorithm complexity
Searching lists is O(n), so we'd need to implement a look-up table to find the last queued command of a specific ID in O(1).

#### Race conditions

At a low level, this makes sense, but how does this interact with the application layer? Say someone issues two "poe reload" commands in rapid succession. This is not that unusual when someone edits the config and saves often. procd sends a `poe reload` on config file changes.

1. `poe reload` is issued
2. first set of packets is queued
3. `poe reload` is issued
4. second set of packets is queued and coalesced with the first.

This might not be problematic from a user's perspective -- they only want the last config applied. It does, however, create a race condition leading to unpredictable behavior. How many packets get coalesced depends on how close the `poe reload`s are in time.

#### Building a bigger picture

You could solve the above by adding explicit coalescing barriers. You could take this one step further, and make the command queue a serialized descriptor of the desired config.

And herein lies the rub: You already have the big picture stored in `struct config`. You just need to serialize it. The next logical step is to serialize it atomically in the first place, and forego the last three points of added complexity,

We've come full circle back to where we started.

svanheule:
Shouldn't a reload cause the command queue to be emptied of old frames anyway? You're parsing new settings, so there's no point in sending config that will soon be replaced. I also find it strange that you're ignoring parts of the config file on reloads... (e.g. global power budget)

mrnuke:
> Shouldn't a reload cause the command queue to be emptied of old frames anyway?

It depends. By the time a `poe reload` comes in, the daemon is most likely past config and sending port queries. If only one port is changed, it makes no sense to interrupt the flow of commands.

There is a(n evil) way to empty out the command queue:

    /etc/init.d/poe restart

> I also find it strange that you're ignoring parts of the config file on reloads... (e.g. global power budget)

Hehe, don't `git blame` me for that.

Hurricos:
While I don't have a practical setup to measure how reducing the number of frames improves reliability, I *am* going to test raw frame counts using

```sh
realtek-poe -d 2>&1 | dd of=/dev/null status=progress bs=1
```

over 90s on my GS1900-24HPv1, just to get an idea of what sort of reduction we're seeing.

Before:
```sh
root@OpenWrt:/# realtek-poe -d 2>&1 | dd of=/dev/null status=progress bs=1
222055 bytes (222 kB, 217 KiB) copied, 90 s, 2.5 kB/s^C
222936+0 records in
222936+0 records out
222936 bytes (223 kB, 218 KiB) copied, 90.7382 s, 2.5 kB/s
```

(edited) After (tested twice):
```sh
root@OpenWrt:~# realtek-poe -d 2>&1 | dd of=/dev/null status=progress bs=1
211891 bytes (212 kB, 207 KiB) copied, 90 s, 2.4 kB/s^C
213024+0 records in
213024+0 records out
213024 bytes (213 kB, 208 KiB) copied, 90.4276 s, 2.4 kB/s
root@OpenWrt:~# realtek-poe -d 2>&1 | dd of=/dev/null status=progress bs=1
211891 bytes (212 kB, 207 KiB) copied, 90 s, 2.4 kB/s^C
213024+0 records in
213024+0 records out
213024 bytes (213 kB, 208 KiB) copied, 90.3878 s, 2.4 kB/s
```

Seeing that this is only a reduction in the number of frames sent during setup, and seeing as setup is finished at around the 4s mark, this is a pretty hefty reduction (close to 40%), It also seems to me that `ubus call poe info` becomes available sooner after starting.

src/main.c
```diff
@@ -666,6 +668,9 @@ poe_port_setup(void)
 {
 	size_t i;
 
+	poe_cmd_port_disconnect_type(PORT_ID_ALL, 2);
```
Hurricos, line 724: I definitely believe we won't need to configure disconnect detection type on a per-port basis, based on commentary from @svanheule's [documentation](https://svanheule.net/switches/software/broadcom_poe_control_protocol#set_disconnect_type).

src/main.c
```diff
@@ -666,6 +668,9 @@ poe_port_setup(void)
...
+	poe_cmd_port_detection_type(PORT_ID_ALL, 3);
```
Hurricos, line 725: The documentation is less certain on this point about the _connect_ detection type setting, but I lack deep knowledge of the 802.3a{f,t} protocol to say, so this is fine.

Hurricos (approved):
Looks good to me. I'm a little concerned that this might complicate your work in #15, but on the other hand, I think we've established that the MCU prefers to get fewer packets.

Hurricos:
@mrnuke Is this a draft, or can I boot it into a regular PR and merge?

src/main.c
mrnuke: That's exactly why I'm not changing the setting itself. Just how we send it to the MCU...

mrnuke:
@Hurricos , the code is complete as is. Keeping as draft -- pending further testing, just in case something really crazy happens.

mrnuke:
@Hurricos, oh, I've missed your [comment](https://github.com/Hurricos/realtek-poe/pull/19#issuecomment-1189211970) about extensive testing. If this didn't go up in flames, then I think we're good to go.

Hurricos:
Let me test the new rebase.

Hurricos:
Test looks good. It has similar issues as before: sometimes data gets "stuck" and doesn't come back:

```txt
                "lan21": {
                        "mode": "PoE",
                        "status": "Delivering power",
                },
```
... as opposed to:

```txt
                "lan21": {
                        "priority": 2,
                        "mode": "PoE",
                        "status": "Delivering power",
                        "consumption": 5.200000
                },
```

This doesn't stop power delivery, and for now, an `/etc/init.d/poe restart` fixes the issue, while (I think) #15 will deal with those instances without intervention.

I'm happy merging this.

Hurricos:
I have tested changing configuration on ports as well; works as expected.

Merging.

PR #20, opened 2022-07-10
Log MCU errors in the system log

mrnuke:
This is a subset of PR #15. It includes everything needed to log error replies, but drops the retry mechanism.

@Hurricos, I'd like to get this merged soon, so that when issues do pop up, we have some log indication of what happened. It doesn't solve any issue -- just logs the relevant bits:

    logread -te realtek-poe

But wait, there's more! Also found a memory leak that I'm confident is fixed by this PR.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -548,6 +548,24 @@ static poe_reply_handler reply_handler[] = {
 	[0x30] = poe_reply_port_power_stats,
 };
 
+static void handle_poe_f0_reply(struct cmd *cmd, uint8_t *reply)
+{
+	const char *reason;
+
+	const char *reasons[] = {
+		[0xd] = "request-incomplete",
+		[0xe] = "request-bad-checksum",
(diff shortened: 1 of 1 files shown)


Hurricos:
The logging for MCU errors is pretty heavy. I notice it ends right as procd's init completes:

```txt
Tue Jul 19 16:21:27 2022 daemon.notice realtek-poe: MCU rejected command: request-bad-checksum
Tue Jul 19 16:21:27 2022 daemon.notice realtek-poe:     CMD:    26 88 14 ff ff ff ff ff ff ff ff ba
Tue Jul 19 16:21:27 2022 daemon.notice realtek-poe:     reply:  fe ff ff ff ff ff ff ff ff ff ff f4
Tue Jul 19 16:21:27 2022 daemon.info procd: - init complete -
```

It was at this point that I realized that the point of this MR is to make these issues visible rather than invisible.

Merging.

PR #22, opened 2022-07-23
realtek-poe: Add support for PoE control over I2C

ghost:
This adds PoE control over I2C instead of the serial
console.

The protocol logic is different from the serial logic,
since an I2C slave cannot initiate transfers. Instead,
after each command (I2C write) sent, the master immediately
fetches the response from the MCU via an I2C read via
a single I2C r/w cycle. The I2C controller will if necessary
wait for the slave to be ready to send its reply in the
I2C read transfer.

There are no further libraries necessary to support I2C, as
only iocl() calls are used on an i2c device in /dev/i2c-*
We therefore always build with i2c support.
For this to work, the kernel needs to be built with
CONFIG_I2C_CHARDEV=y
in order for the i2c character devices to be present in
/dev.

I2C instead of UART is selected by simply putting e.g. a
	option i2c_bus	'3'
in the global section of the poe configuration file,
where '3' gives the number of the I2C character device.

Optionally, an i2c_addr can be specified, which
defaults to 0x20 as for the address in the only 2 known
PoE devices with I2C.

Signed-off-by: Birger Koblitz <git@birger-koblitz.de>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -9,7 +9,9 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <getopt.h>
-
+#include <linux/i2c.h>
+#include <linux/i2c-dev.h>
+#include <sys/ioctl.h>
 #include <libubox/ustream.h>
 #include <libubox/uloop.h>
 #include <libubox/list.h>
@@ -41,6 +43,8 @@ struct port_config {
 struct config {
 	float budget;
 	float budget_guard;
+	int i2c_bus;
+	int i2c_addr;
 
 	unsigned int port_count;
 	uint8_t pse_id_set_budget_mask;
@@ -71,6 +75,7 @@ struct cmd {
 };
 
 static struct ustream_fd stream;
+static int i2c_file;
 static LIST_HEAD(cmd_pending);
 static unsigned char cmd_seq;
 static struct state state;
@@ -81,6 +86,8 @@ static struct config config = {
 	.budget_guard = 7,
 	.pse_id_set_budget_mask = 0x01,
 };
+
+static int poe_reply_consume(unsigned char *reply);
 
 static uint16_t read16_be(uint8_t *raw)
 {
@@ -130,15 +137,26 @@ static void load_port_config(struct uci_context *uci, struct uci_section *s)
 
 static void load_global_config(struct uci_context *uci, struct uci_section *s)
 {
-	const char *budget, *guardband;
+	const char *budget, *guardband, *i2c_bus, *i2c_addr;
 
 	budget = uci_lookup_option_string(uci, s, "budget");
 	guardband = uci_lookup_option_string(uci, s, "guard");
+	i2c_bus = uci_lookup_option_string(uci, s, "i2c_bus");
+	i2c_addr = uci_lookup_option_string(uci, s, "i2c_addr");
 
 	config.budget = budget ? strtof(budget, NULL) : 31.0;
 	config.budget_guard = config.budget / 10;
 	if (guardband)
 		config.budget_guard = strtof(guardband, NULL);
+	if (i2c_bus) {
+		config.i2c_bus = strtol(i2c_bus, NULL, 10);
+		if (i2c_addr)
+			config.i2c_addr = strtol(i2c_addr, NULL, 10);
+		else
+			config.i2c_addr = 0x20; // I2C address on both known devices
+	} else {
+		config.i2c_bus = -1;
+	}
 }
 
 static void
@@ -225,6 +243,35 @@ poe_cmd_send(struct cmd *cmd)
 }
 
 static int
+poe_i2c_cmd_send(struct cmd *cmd, unsigned char *reply, int len)
+{
+	int nmsgs_sent;
+        struct i2c_msg msgs[2];
+	struct i2c_rdwr_ioctl_data rdwr;
+
+	log_packet(LOG_DEBUG, "I2C TX ->", cmd->cmd);
+
+	msgs[0].addr = config.i2c_addr;
+	msgs[0].buf = cmd->cmd;
+	msgs[0].flags = 0;
+	msgs[0].len = 12;
+
+	msgs[1].addr = config.i2c_addr;
+	msgs[1].buf = reply;
+	msgs[1].flags = I2C_M_RD;
+	msgs[1].len = len;
+
+	rdwr.msgs = msgs;
+	rdwr.nmsgs = 2;
+
+	nmsgs_sent = ioctl(i2c_file, I2C_RDWR, &rdwr);
+	if (nmsgs_sent != 2)
+		ULOG_NOTE("Could not complete I2C send/receive, messages sent: %d\n", nmsgs_sent);
+
+	return 0;
+}
+
+static int
 poe_cmd_next(void)
 {
 	struct cmd *cmd;
@@ -242,6 +289,7 @@ poe_cmd_queue(unsigned char *_cmd, int len)
 {
 	int i, empty = list_empty(&cmd_pending);
 	struct cmd *cmd = malloc(sizeof(*cmd));
+	unsigned char reply[12];
 
 	memset(cmd, 0, sizeof(*cmd));
 	memset(cmd->cmd, 0xff, 12);
@@ -256,8 +304,13 @@ poe_cmd_queue(unsigned char *_cmd, int len)
 
 	list_add_tail(&cmd->list, &cmd_pending);
 
-	if (empty)
+	if (!empty)
+		return 0;
+	if (config.i2c_bus < 0)
 		return poe_cmd_send(cmd);
+
+	poe_i2c_cmd_send(cmd, reply, sizeof(reply));
+	poe_reply_consume(reply);
 
 	return 0;
 }
@@ -603,12 +656,12 @@ poe_reply_consume(unsigned char *reply)
 	free(cmd);
 
 	if ((reply[0] != cmd_id) || (reply[0] > ARRAY_SIZE(reply_handler))) {
-		ULOG_DBG("received reply with bad command id\n");
+		ULOG_INFO("received reply with bad command id\n");
 		return -1;
 	}
 
 	if (reply[1] != cmd_seq) {
-		ULOG_DBG("received reply with bad sequence number\n");
+		ULOG_INFO("received reply with bad sequence number\n");
 		return -1;
 	}
 
@@ -677,6 +730,37 @@ poe_stream_open(char *dev, struct ustream_fd *s, speed_t speed)
 	tcflush(tty, TCIFLUSH);
 
 	return 0;
+}
+
+static int
+i2c_addr_open(int i2c_bus, char *filename, size_t size)
+{
+	int file, len;
+
+	len = snprintf(filename, size, "/dev/i2c-%d", i2c_bus);
+        if (len >= (int)size) {
+		ULOG_ERR("%s: path truncated\n", filename);
+                return -EOVERFLOW;
+        }
+        file = open(filename, O_RDWR);
+
+	if (file < 0)
+		ULOG_ERR("Error: Could not open file `%s': %s\n", filename, strerror(errno));
+
+	// TODO: check needed I2C functionality
+	return file;
+}
+
+static int
+i2c_set_slave_addr(int file, int address)
+{
+        if (ioctl(file, I2C_SLAVE, address) < 0) {
+                ULOG_ERR("Error: Could not set address to 0x%02x: %s\n",
+                         address, strerror(errno));
+                return -errno;
+        }
+
+        return 0;
 }
 
 static int poet_setup(const struct port_config *ports, size_t num_ports)
@@ -910,6 +994,7 @@ int
 main(int argc, char ** argv)
 {
 	int ch;
+	char i2c_dev_filename[20];
 
 	struct uloop_timeout state_timeout = {
 		.cb = state_timeout_cb,
@@ -936,9 +1021,17 @@ main(int argc, char ** argv)
 	uloop_init();
 	ubus_auto_connect(&conn);
 
-	if (poe_stream_open("/dev/ttyS1", &stream, B19200) < 0)
-		return -1;
-
+	if (config.i2c_bus >= 0) {
+		i2c_file = i2c_addr_open(config.i2c_bus, i2c_dev_filename, sizeof(i2c_dev_filename));
+		if (i2c_file < 0)
+			return -1;
+		ULOG_INFO("Using I2C bus #%d: %s\n", config.i2c_bus, i2c_dev_filename);
+		ULOG_INFO("Setting I2C slave address to %x\n", config.i2c_addr);
+		i2c_set_slave_addr(i2c_file, config.i2c_addr);
+	} else {
+		if (poe_stream_open("/dev/ttyS1", &stream, B19200) < 0)
+			return -1;
+	}
 
 	poe_initial_setup();
 	uloop_timeout_set(&state_timeout, 1000);
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -16,8 +16,9 @@ FIND_LIBRARY(uci NAMES uci)
 FIND_LIBRARY(ubox NAMES ubox)
 FIND_PATH(uci_include_dir NAMES uci.h)
 FIND_PATH(ubus_include_dir NAMES libubus.h)
+FIND_PATH(i2c_include_dir NAMES i2c.h)
 FIND_PATH(ubox_include_dir NAMES libubox/usock.h)
-INCLUDE_DIRECTORIES(${ubox_include_dir} ${ubus_include_dir} ${uci_include_dir})
+INCLUDE_DIRECTORIES(${ubox_include_dir} ${ubus_include_dir} ${uci_include_dir} $(i2c_include_dir))
 
 ADD_EXECUTABLE(realtek-poe ${SOURCES})
 


mrnuke:
Hi @bkobl 

The first thing I see is the need to separate out the UART and I2C paths. Any part of the code that is not I2C or UART specific should not have to care if the MCU is on UART or I2C. `if()/else()` inside the queue handler goes the opposite direction.

Second thing I see is the I2C ioctl handles both the packet and and receive side, in a blocking manner. realtek-poe uses an asynchronous model where it fires off data and waits for things to happen. That way, we can respond to ubus requests while waiting for MCU packets. It's really cool. As a request happens, we process it, and send the response, all while the MCU communication is in progress.

With the way the I2C code is added, we're blocking program execution for the entire duration of the transfer. How long does an I2C round-trip take. How long do we block the ubus for the setup packets, and how long for each subsequent port poll? It seems that the current approach would make realtek-poe very unresponsive.

[] Is there a non-blocking I2C kernel call that we can use instead?
[] How de we integrate blocking calls into the asynchronous design of realtek-poe?

ghost:
> The first thing I see is the need to separate out the UART and I2C paths. Any part of the code that is not I2C or UART specific should not have to care if the MCU is on UART or I2C. `if()/else()` inside the queue handler goes the opposite direction.

This is due to I2C being a synchronous protocol, whereas UART is asynchonous. There is necessarily going to be something in the queue handler that has to distinguish between the two. One could define 2 different modes: synchonous and asynchronous processing, and then do the if()s on that condition. The asynchronous code path would then do the UART transfers and any future 

> Second thing I see is the I2C ioctl handles both the packet and and receive side, in a blocking manner. realtek-poe uses an asynchronous model where it fires off data and waits for things to happen.

As already stated in the PR description, this is how it needs to be. I2C is a synchonous protocol (all transfers are initiated by master), while UART is asynchonous (slave can initiate transfer). The blocking is limited to < 60ms until even a started transfer times out. Note that UART is also blocking at that level of time, namely during the transfer itself, since the kernel is not preempting transfers. Also the I2C read is not really blocking until a reply is received, since the master initiates the transfer. A typical transfer takes 41 ms and if the slave does not reply within 10ms of the beginning of the transfer, the transfer times out, this causes all ioctl() to return within 10 to less than 60ms. On the time-scale of ubus, this is not really blocking anything.

I have also tested error conditions by blocking the I2C bus in hardware (just touch a strong pull-down to SDA) the code just happily ignores this, times out and retries or gets wrong checksums and continues (there is no error handling of wrong checksums in the code so far it seems):
```
realtek-poe: I2C TX -> 03 2c 01 01 ff ff ff ff ff ff ff 2a
realtek-poe: RX <- ff ff ff ff ff ff ff ff ff ff ff ff
realtek-poe: received reply with bad checksum
realtek-poe: I2C TX -> 00 2d 02 01 ff ff ff ff ff ff ff 29
realtek-poe: RX <- 00 2d 02 00 ff ff ff ff ff ff ff 28
```

[] Is there a non-blocking I2C kernel call that we can use instead?
No, I have searched the net and kernel for this. I2C calls are guaranteed to return in such a short amount of time it would anyway not make sense. The requirement to answer quickly is part of the I2C protocol in order not to block the I2C bus for other devices. This is different for a serial bus with 2 equal participants.
[] How de we integrate blocking calls into the asynchronous design of realtek-poe?
I don't think there is a need to. Even when blocking the I2C bus completely, these calls do not block, but time out immediately.

mrnuke:
I think you're confusing asynchronous link layer (not sharing the same clock) with asynchronous software design. I don't care what the underlying hardware does. I care that the application is responsive. The typical 41ms transfer turns to several seconds of deadlock whenever firing a burst of requests.

If you have the main uloop toggle a gpio every 100ms, what happens when I2C transactions start flowing?

src/main.c
```diff
@@ -41,6 +43,8 @@ struct port_config {
 struct config {
 	float budget;
 	float budget_guard;
+	int i2c_bus;
```
Hurricos, line 46: Which boards use I2C to communicate to the MCU for the [Broadcom-based PoE command frame implementation](https://svanheule.net/switches/software/broadcom_poe_control_protocol)? I ask because I wonder whether we can fetch the address and information at runtime from the devicetree.

src/main.c
```diff
@@ -130,15 +137,26 @@ static void load_port_config(struct uci_context *uci, struct uci_section *s)
 
 static void load_global_config(struct uci_context *uci, struct uci_section *s)
 {
-	const char *budget, *guardband;
+	const char *budget, *guardband, *i2c_bus, *i2c_addr;
```
Hurricos, line 140: As with my previous comment, if we can pull these from the device tree, there'd be no need to read this data in from  configuration file.

src/main.c
```diff
@@ -224,6 +242,35 @@ poe_cmd_send(struct cmd *cmd)
 	return 0;
 }
 
+static int
+poe_i2c_cmd_send(struct cmd *cmd, unsigned char *reply, int len)
```
Hurricos, line 246: Seems like we should separate both the serial and I2C implementations out, so that main.c can be purely about sending frames, not frames *through a specific medium*.

src/main.c
```diff
@@ -603,12 +656,12 @@ poe_reply_consume(unsigned char *reply)
 	free(cmd);
 
 	if ((reply[0] != cmd_id) || (reply[0] > ARRAY_SIZE(reply_handler))) {
-		ULOG_DBG("received reply with bad command id\n");
+		ULOG_INFO("received reply with bad command id\n");
```
Hurricos, line 659: Hmm, what's the reasoning behind escalating this from DBG to INFO?

src/main.c
```diff
@@ -679,6 +732,37 @@ poe_stream_open(char *dev, struct ustream_fd *s, speed_t speed)
 	return 0;
 }
 
+static int
+i2c_addr_open(int i2c_bus, char *filename, size_t size)
```
Hurricos, line 736: As with the other reply, I know `main.c` is currently already messy with medium-specific implementations, but I'd prefer we separated media-handling out. We're not trying to clone `poemgr`, but this is how `poemgr` does it: note that `poemgr.c` contains no media-handling-specific headers, while the specific [PD IC file `pd69104.c`](https://github.com/blocktrron/poemgr/blob/master/pd69104.c) does.

Hurricos (changes requested):
Added some comments. I know this is a big ask, but I think medium-handling should be separated. We may want to work on that as part of a separate issue before we merge I2C support.

This is a brilliant find, BTW. I guess I'm not that surprised that the same protocol is used when the MCU is attached via I2C, Broadcom has a habit of maintaining its own standards for things.

src/main.c
svanheule: Added a column toe the [PoE management](https://svanheule.net/switches/software/poe_management) overview page to keep track of this info. I have a GS1920-24HP, but that's currently in use. To get the device info, realtek-poe could scan for certain compatibles. That means the MCU needs it's own node in the device tree, which is currently not the case.

ghost:
> The typical 41ms transfer turns to several seconds of deadlock whenever firing a burst of requests.
No. This is not true. The _entire_ transfer cycle takes maximum 60 milliseconds. In this sychronous design, there is only one request at a time. There are no longer bursts of requests, because they are not queued. And even then, if you would be counting the time for a complete "burst of requests", the longest I can see is 350ms at startup, and this can definitely be interrupted in the middle, i.e. the application stays responsive. The monitoring requests take about 200ms for the entire burst.

If 350ms are too long, the I2C speed can be raised. I went for something very conservative, but the OEM software does things 8 times faster, so an entire burst would be about 50ms.

> I think you're confusing asynchronous link layer (not sharing the same clock) with asynchronous software design. I don't care what the underlying hardware does. I care that the application is responsive. The typical 41ms transfer turns to several seconds of deadlock whenever firing a burst of requests.
> 
> If you have the main uloop toggle a gpio every 100ms, what happens when I2C transactions start flowing?
The uloop does not toggle GPIO, it toggles complete I2C transfers. The toggling of GPIOs happens in the GPIO-I2C driver at 100kHz.

src/main.c
ghost: I would prefer that, too. One possibility to detect this, would also be to give the I2C bus in the .dts a nice name, such as "i2c-poe", these names show up in the /dev and /sys filesystems. It is further possible to probe that bus, in order to detect the right device address. Probing could be done within realtek-poe, there would not be the need to have additional tools outside.

src/main.c
ghost: Indeed.

src/main.c
ghost: Yes, we should have a backend. It would also solve the issue of the synchronous/asynchronous behaviour. If the backend would be in a separate thread, we could have everything asynchronous.

src/main.c
ghost: This was during debugging, I wanted to be sure that I see if something really bad happens during transfers. But this is not necessary, I shall fix this.
Screenshot from 2022-07-27 20-51-33

The I2C protocol produces a large number of failed transfers, about 1% fail. Compared to the OEM I2C-GPIO driver (which has optimizations specific to the RTL SoCs), the timing of the transfers look horrible, see the above image. In the above example, the reply checksum fails, probably because the I2C timing confuses the MCU, and it would be sufficient to retry. I am just now wondering how this can be done better in `poe_reply_consume()`:

```
	cmd = list_first_entry(&cmd_pending, struct cmd, list);
	list_del(&cmd->list);
	cmd_id = cmd->cmd[0];
	cmd_seq = cmd->cmd[1];

	for (i = 0; i < 11; i++)
		sum += reply[i];

	if (reply[11] != sum) {
		ULOG_DBG("received reply with bad checksum\n");
		free(cmd);
		return -1;
	}
```
The command needs to be resent and not discarded when the checksumming fails. If commands during the startup fail, then PoE fails to work.

Hurricos: There's some logic for this being prepared on the UART side, see #15. Since the logic has been developed for the other medium, it'd be quite nice if they could share code-paths completely.

src/main.c
Hurricos: I'm going to look into .dts compatible bindings that we could use (or create!) to have the kernel begin buffering i2c reading for us. I doubt one exists for our specific case, but i2c is ubiquitous, so there's probably *something* out there.

Edit: by buffering, I mean polling. I know i2c is necessarily synchronous, but if we could ask the kernel to make it async for us using software (as it does with hardware by interacting with ns16650-compatible hardware on the UART side), that would be very convenient. It makes sense, too, seeing as we're already rooting around in the device tree to avoid configuring this manually.

src/main.c
mrnuke: @bkobl, that transfer looks way faster than the capture you shared on the forum. Do the I2C lines go over opto-isolators, or isolation transformers? If that's the case, you might get signal integrity issues if you drive the bus too fast. It looks like the MCU is trying to do clock stretching to slow down the transfer.

Also, reply ff means `not-ready`. That also happens on UART when sending requests in fast succession. I could be the same root cause as #10 .

src/main.c
```diff
@@ -256,9 +304,14 @@ poe_cmd_queue(unsigned char *_cmd, int len)
...
 	list_add_tail(&cmd->list, &cmd_pending);
 
-	if (empty)
+	if (!empty)
+		return 0;
+	if (config.i2c_bus < 0)
 		return poe_cmd_send(cmd);
 
+	poe_i2c_cmd_send(cmd, reply, sizeof(reply));
+	poe_reply_consume(reply);
```
Hurricos, line 313: Are we SURE we cannot use `ustream_fd_init` (ustream / uloop) to consume instead?

src/main.c
Hurricos: Similar to what @bkobl mentioned:

> No, I have searched the net and kernel for this. I2C calls are guaranteed to return in such a short amount of time it would anyway not make sense. The requirement to answer quickly is part of the I2C protocol in order not to block the I2C bus for other devices. This is different for a serial bus with 2 equal participants.

... there is no fancy binding out there to have the kernel call-back on code. There would be no point to having such a thing: a driver interacting with i2c is either entirely devoted to that i2c device, or can be written in a threaded manner so that blocking on an i2c read is no big deal.

src/main.c
```diff
@@ -130,15 +137,26 @@ static void load_port_config(struct uci_context *uci, struct uci_section *s)
...
 	config.budget = budget ? strtof(budget, NULL) : 31.0;
 	config.budget_guard = config.budget / 10;
 	if (guardband)
 		config.budget_guard = strtof(guardband, NULL);
+	if (i2c_bus) {
+		config.i2c_bus = strtol(i2c_bus, NULL, 10);
+		if (i2c_addr)
+			config.i2c_addr = strtol(i2c_addr, NULL, 10);
+		else
+			config.i2c_addr = 0x20; // I2C address on both known devices
```
Hurricos, line 156: Please drop this assumption from code: I would prefer the code break than be board-specific.

src/main.c
```diff
@@ -256,9 +304,14 @@ poe_cmd_queue(unsigned char *_cmd, int len)
 
 	list_add_tail(&cmd->list, &cmd_pending);
 
-	if (empty)
+	if (!empty)
+		return 0;
+	if (config.i2c_bus < 0)
```
Hurricos, line 309: The way that this case statement has been rewritten makes it very difficult to suss out what's special about i2c vs. uart reply handling.

After a few re-readings, I understand that

- this does not change control flow for uart, and
- for i2c, as long as there are no items in `cmd_pending` at the start of `poe_cmd_queue`, then we continue and invoke `poe_i2c_cmd_send`.

`poe_cmd_queue` is the only place where `poe_i2c_cmd_send`. This means that the i2c implementation has hijacked the 'queue': it will only carry the command being sent at any point in time. Should `cmd_pending` ever contain anything else, this will never complete, as `poe_i2c_cmd_send` can never process.

You haven't just implemented i2c more synchronously, you've also watered down the functionality of the command queue. Even if this patch works, it means we cannot rely on `cmd_pending` to work as a queue later down the line (e.g. if we want to enqueue a packet but not process it right now, it'll break i2c).

Please either re-implement i2c through ustream, or design / wait for an implementation of reply queueing that doesn't use it. I understand that it will be less efficient (because `ustream` `read()`s instead of using the `I2C_RDWR` ioctl). I'll be trying to work out the latter option, but my understanding of C threading is abysmal.

Hurricos (changes requested):

src/main.c
Hurricos: > Which boards use I2C to communicate to the MCU

@svanheule @bkobl: thanks for the [I2C column](https://svanheule.net/switches/software/poe_management), but I don't see any boards using I2C that have device trees in OpenWrt yet. I'm probably mistaken -- let me know about that so I can take a swing at adding this MCU to their device trees.

src/main.c
Hurricos: Answered myself -- we absolutely could read using ustream, but, downsides:
- ustream would still block on read (it cannot poll() to skip i2c)
- we'd be losing out on the efficiency of the I2C_RDWR ioctl, and
- (as per @mrnuke's analogy) we'd be streaminfying the I2C packets and then re-packetizing them: the I2C device will already be yielding 12-byte packets when it has anything to give.

src/main.c
svanheule: No, I don't think you're mistaken. Most devices use the UART-variant, and neither of the two other devices from the table are in OpenWrt. My GS1920-24HP has been back at work since a while, as I sort of already had enough devices at home to play with (that have more useful bootloaders).

Hurricos:
Converting to a draft. I can't merge this until OpenWrt hardware has been ported for me to test this on.

Hurricos:
Lots of thoughts:

- We heard from jow today that ustream is already very heavy-handed in its application to uart. uloop with a handmade uloop_fd to create callbacks would be preferable.
- I've added my reasons for not liking how this is implemented; particularly [here](https://github.com/Hurricos/realtek-poe/pull/22#discussion_r934023409)
- @mrnuke has more than one reason that tx'ing i2c using uloop or uloop-derived mechanisms (e.g. ustream) to poll is not a straightforward choice
  - Using a C11 thread to handle i2c transactions and avoid waits is in the cards, but ...
- Most importantly, until OpenWrt hardware is available for us to test other options on, we can't merge this.

Give us a ping when one of these devices gets OpenWrt support merged.

src/main.c
Hurricos: Not ustream. jow points out that that's a horrible idea; use e.g.

```txt
 maybe do the required EAGAIN / EINTR nonblock IO repeat bilerplate around
 struct my_stuff { struct uloop_fd ufd; other stuff };
 struct my_stuff *blah = calloc(1, sizeof(struct my_stuff));
 my_stuff->ufd.fd = socket(...); // or open() or whatever you gyus do there
 my_stuff->...; // init other stuff you might need
 my_stuff->ufd.cb = my_read_callback;
 uloop_fd_add(&my_stuff->ufd, ULOOP_READ);
 in your callback:
 struct my_stuff *blah = container_off(ufd /* the uloop_fd pointer you get as 1st arg in the callback */, struct my_stuff, ufd);
 *container_of() 
 ```

File: src/main.c (hunk 4 above)
Comment (ghost) [reply src/main.c:659]: I investigated different clock speeds a bit more. Slowing down the bus to 10kHz isn't making anything better. Instead it is becoming obvious that there is an issue with the timing of the I2C bitbanging
Screenshot from 2022-08-01 08-59-41

It looks like every 2ms there is a pause, or maybe the clock is slowed down. This is an RTL839x system with a second VPE and a second clock which is used as a timer (i.e. it can be used to make the CPU sleep and then have an IRQ triggered to wake it up, including switching the VPE context at that point). This is much more complex than the situation on the RTL838x. Althought the transfers look horrible in the trace, they actually succeed. I don't think there is clock slowing on the side of the MCU. This is usually used when there is an issue with the I2C peripherial, not to indicate that the MCU has not finished computations.

The problem with the all FF response seems indeed to come from the MCU not being ready, it seems to happen within a rapid sequency of commands, not asking for monitor data. So I'll do some retrying, you said there was already some logic.

File: src/main.c (hunk 7 above)
Comment (ghost) [reply src/main.c:156]: I'll follow you suggestion here. But I don't believe this is going to be board specific, since there is some generic SDK behind that. Anyway, we could also do auto-detection here.

File: src/main.c (hunk 8 above)
Comment (ghost) [reply src/main.c:309]: I would prefer to have a threaded implementation that entirely hides I2C and UART behind it. They would both be synchronous, and the thread would provide the asynchronicity. There would be a thread for UART and one for I2C depending on what is needed. We could also handle command repeats in the thread so that the queueing is cleaner. The queue would be the interface between the asynchronous front-end and a synchronous backend which also handles some of the error cases, like command repeats. After all it would know better about timing.
I have done some heavy pthread stuff in the past, but I have not touched it since more than a decade, and I would probably need to start again at more-or-less zero.

But first I need to get the PoE going at all, the failing commands lead to the power not coming up properly.

File: src/main.c (hunk 1 above)
Comment (ghost) [reply src/main.c:46]: The Netgear GS728TPv2 support is ready to be included but needs some other PRs to be merged first. It even has a working USB port! But frankly no one will use this device unless PoE works, so while PoE does not work I am not in a hurry. Personally, I need the GS728TPv2 at the end of the Summer to become the core switch at my Summer place driving plenty of sensors and WiFI repeaters over PoE. Until then it needs to work, but it also limits the time I can play around with it.

File: src/main.c (hunk 8 above)
Comment (mrnuke) [reply src/main.c:309]: For the failing commands, please try to rebase on #15 .

I have some ideas where to snip the interface. As long as the backend gets its commands from the `cmd_pending` queue, and puts the replies in via `poe_reply_consume()`, that's the basis of the interface. Whether the backend runs in a thread, or as part of uloop() with epoll(), it doesn't matter. So there would be no need to do a thread for UART.

I'll try to get a basic split of UART code in the next few days. There's some context tracking and other boilerplate that I might need to implement.

Comment (ghost) [reply src/main.c:309]: The error handling comes too late for the I2C code and it also does not apply to the checksum errors. If there is a checksum error, the code never gets into `handle_poe_f0_reply()` in `poe_reply_consume()`.
What works for me is the following:
```
static int
poe_cmd_queue(unsigned char *_cmd, int len)
{
	int i, empty = list_empty(&cmd_pending);
	struct cmd *cmd = malloc(sizeof(*cmd));
	unsigned char reply[12];
	unsigned char sum;
	int retries = 10;

	memset(cmd, 0, sizeof(*cmd));
	memset(cmd->cmd, 0xff, 12);
	memcpy(cmd->cmd, _cmd, len);

	cmd_seq++;
	cmd->cmd[1] = cmd_seq;
	cmd->cmd[11] = 0;

	for (i = 0; i < 11; i++)
		cmd->cmd[11] += cmd->cmd[i];

	list_add_tail(&cmd->list, &cmd_pending);

	if (!empty)
		return 0;
	if (config.i2c_bus < 0)
		return poe_cmd_send(cmd);

	do {
		poe_i2c_cmd_send(cmd, reply, sizeof(reply));
		sum = 0;
		for (i = 0; i < 11; i++)
			sum += reply[i];
		if (reply[11] != sum) {
			ULOG_DBG("received reply with bad checksum, RETRY\n");
		} else {
			break;
		}
		usleep(10000);
	} while (--retries);

 	poe_reply_consume(reply);

	return 0;
}
```
That's why I would suggest to handle checksum errors with a simple retry in a thread that handles the top of the command queue.

Using the above code the all FF responses are no longer an issue. It seems indeed as if the MCU is not ready and repeating the command after 10ms seems always to succeed.

mrnuke: Ohh, the all F's reply has an invalid checksum on I2C. Fun!

Feel free to develop the I2C path so that you can power on the ports. The goal is to obtain the knowledge on how to drive the MCU to make ports work. The final code will likely change many times, as we split the backends.

Neustradamus:
@Hurricos: Now the PR author account is a ghost, what is the situation now for this PR?

mrnuke:
> @Hurricos: Now the PR author account is a ghost, what is the situation now for this PR?

It's haunted.

Issue #11, opened 2022-06-28
Remove bash-style function declarations from C code


mrnuke:
In shell scripts, functions do not have a return type

```
do_shell_stuff() {
}
```
### Bad C code
On the other hand, in C, the return type of a function is crucial to the understanding of the function and surrounding code. The function is non severable from its return type. Thus lines like the following are non-sensical:

```
static int
do_c_stuff(void()
{
}
```

While I suspect there is a desire to write functions this way for consistency with bash code, I find this to be stupid. It detracts readability because:
1. It increases the number of lines of code
2. It requires more cognitive effort to understand the function

### Preferred C code

It is always preferable do declare the function in a single line. If that can be done, it is the only way to do it.

```
static int do_c_stuff(void)
{
}
```

When this does not fit on a single line (80 characters according to the kernel coding guidelines), it is okay to mode the **static** and/or **inline** keyword on a separate line. In any case the return type must be on the same line as the function.
```
static inline
struct long_name_struct *do_long_function_name(int lots, size_t of, char arguments)
{
}
```

### Issue

A lot of code in realtek-poe uses the "Bad C code" style. This must be refactored.

PR #24, opened 2022-08-02
realtek-poe: add ubus port power control

bmork:
Allow power to be dynamically disabled/enabled on individual
ports, similar to the ancient lua implementation

Note that "port" parameter now takes the configured port name
instead of the one-based index used by the lua code.

Signed-off-by: Bjørn Mork <bjorn@mork.no>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -880,10 +880,40 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
 	return UBUS_STATUS_OK;
 }
 
+static const struct blobmsg_policy ubus_poe_manage_policy[] = {
+	{ "port", BLOBMSG_TYPE_STRING },
+	{ "enable", BLOBMSG_TYPE_BOOL },
+};
+
+static int ubus_poe_manage_cb(struct ubus_context *ctx, struct ubus_object *obj,
+			      struct ubus_request_data *req, const char *method,
+			      struct blob_attr *msg)
+{
+	struct blob_attr *tb[ARRAY_SIZE(ubus_poe_manage_policy)];
+	const struct port_config *port;
+	const char *port_name;
+	size_t i;
+
+	blobmsg_parse(ubus_poe_manage_policy,
+		      ARRAY_SIZE(ubus_poe_manage_policy),
+		      tb, blob_data(msg), blob_len(msg));
+	if (!tb[0] || !tb[1])
+		return UBUS_STATUS_INVALID_ARGUMENT;
+
+	port_name = blobmsg_get_string(tb[0]);
+	for (i = 0; i < config.port_count; i++) {
+		port = &config.ports[i];
+		if (port->enable && !strcmp(port_name, port->name))
+			return poe_cmd_port_enable(i, blobmsg_get_bool(tb[1]));
+	}
+	return UBUS_STATUS_INVALID_ARGUMENT;
+}
+
 static const struct ubus_method ubus_poe_methods[] = {
 	UBUS_METHOD_NOARG("info", ubus_poe_info_cb),
 	UBUS_METHOD_NOARG("reload", ubus_poe_reload_cb),
 	UBUS_METHOD("sendframe", ubus_poe_sendframe_cb, ubus_poe_sendframe_policy),
+	UBUS_METHOD("manage", ubus_poe_manage_cb, ubus_poe_manage_policy),
 };
 
 static struct ubus_object_type ubus_poe_object_type =


src/main.c
```diff
@@ -880,10 +880,42 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+		      tb, blob_data(msg), blob_len(msg));
+	if (!tb[0] || !tb[1])
+		return UBUS_STATUS_INVALID_ARGUMENT;
+
+	port = blobmsg_get_string(tb[0]);
+	for (i = 0; i < config.port_count; i++) {
+		if (!config.ports[i].valid || strncmp(port, config.ports[i].name, sizeof(config.ports[i].name)))
+			continue;
+
+		poe_cmd_port_enable(i, !!blobmsg_get_bool(tb[1]));
```
mrnuke: `blobmsg_get_bool` returns a bool so why the `!!` ?

src/main.c
```diff
@@ -880,10 +880,42 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+
+	blobmsg_parse(ubus_poe_port_policy,
+		      ARRAY_SIZE(ubus_poe_port_policy),
+		      tb, blob_data(msg), blob_len(msg));
+	if (!tb[0] || !tb[1])
+		return UBUS_STATUS_INVALID_ARGUMENT;
+
+	port = blobmsg_get_string(tb[0]);
+	for (i = 0; i < config.port_count; i++) {
+		if (!config.ports[i].valid || strncmp(port, config.ports[i].name, sizeof(config.ports[i].name)))
```
mrnuke: Your reviewer is an evil old-school maggot who likes to keep lines capped at 80 characters. Now, on to the actual points:

1. Do we want to power-on explicitly disabled ports? They have not been configured. Trying to turn them on has undefined results.
2. Isn't `strlen(name)` less than `sizeof(name)` ? Is there a value to str**n**cmp(), when both strings are guaranteed to be null-terminated?
3. Dereferencing an array multiple times on a line. You might want to have a pointer to the element, and dereference it directly.:
 ```
port = &config[i];

if (!port->enabled || strcmp(port_name, port->name))
```

File: src/main.c [hunk 3]
```diff
@@ -880,10 +880,42 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+	{ "enable", BLOBMSG_TYPE_BOOL },
+};
+
+static int
+ubus_poe_port_cb(struct ubus_context *ctx, struct ubus_object *obj,
+		 struct ubus_request_data *req, const char *method,
+		 struct blob_attr *msg)
+{
+	struct blob_attr *tb[ARRAY_SIZE(ubus_poe_port_policy)];
+	char *port;
```
mrnuke: Not modifying the string `port`, and this also represents the name of the port, not the port itself. So should probably be advertised as `const char *port_name`.

src/main.c
```diff
@@ -880,10 +880,42 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
 	return UBUS_STATUS_OK;
 }
 
+static const struct blobmsg_policy ubus_poe_port_policy[] = {
+	{ "port", BLOBMSG_TYPE_STRING },
+	{ "enable", BLOBMSG_TYPE_BOOL },
+};
+
+static int
+ubus_poe_port_cb(struct ubus_context *ctx, struct ubus_object *obj,
```
mrnuke: Please see #11

src/main.c
```diff
@@ -880,10 +880,42 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+		 struct ubus_request_data *req, const char *method,
+		 struct blob_attr *msg)
+{
+	struct blob_attr *tb[ARRAY_SIZE(ubus_poe_port_policy)];
+	char *port;
+	size_t i;
```
mrnuke, line 895: Bonus points for nailing `size_t` instead of int :)

src/main.c
```diff
@@ -880,10 +880,42 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+
+	port = blobmsg_get_string(tb[0]);
+	for (i = 0; i < config.port_count; i++) {
+		if (!config.ports[i].valid || strncmp(port, config.ports[i].name, sizeof(config.ports[i].name)))
+			continue;
+
+		poe_cmd_port_enable(i, !!blobmsg_get_bool(tb[1]));
+		return UBUS_STATUS_OK;
+	}
+	return UBUS_STATUS_NOT_FOUND;
```
mrnuke: This isn't wrong, but this logic seems odd because the expected exit path is on a branch:
```
if (okay()) {
	do_good_things();
	return SUCCESS;
}

return FAIL;
```
Is there a good way to reorganize the flow as the following?
```
if (!okay)
	return FAIL;

do_good_things();
return SUCCESS;
```

src/main.c
```diff
@@ -880,10 +880,42 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+ubus_poe_port_cb(struct ubus_context *ctx, struct ubus_object *obj,
+		 struct ubus_request_data *req, const char *method,
+		 struct blob_attr *msg)
+{
+	struct blob_attr *tb[ARRAY_SIZE(ubus_poe_port_policy)];
+	char *port;
+	size_t i;
+
+	blobmsg_parse(ubus_poe_port_policy,
+		      ARRAY_SIZE(ubus_poe_port_policy),
```
mrnuke: Looks like `blobmsg_parse(ubus_poe_port_policy, ARRAY_SIZE(ubus_poe_port_policy),` will fit on a single line.

mrnuke (changes requested):
We really need a coherent interface over ubus. My concern is that if we add things piecemeal, the result will not be coherent. I'm okay with adding required functionality, with the understanding that the ubus interface will change.

```
ubus call poe port {"port" : "lan42", "enable" : no_senor}
```
Is that the correct form? We need to document this **somewhere**.

EDIT: The ubus command name should contain a verb describing the action to be taken (for example port_**enable**_unlimited_power).

src/main.c
bmork: > We really need a coherent interface over ubus. My concern is that if we add things piecemeal, the result will not be coherent. I'm okay with adding required functionality, with the understanding that the ubus interface will change.

That's where the initial attempts to add this daemon stopped....

Yes, we should have a coherent, well-documented, interface supported on more than one hardware platform.  But no one knows how to get there.  Or, well, at least I don't-

> ```
> ubus call poe port {"port" : "lan42", "enable" : no_senor}
> ```
> 
> Is that the correct form? We need to document this **somewhere**.

I don't think "no_senor" will work as a boolean, but other than that - yes.  

> EDIT: The ubus command name should contain a verb describing the action to be taken (for example port_**enable**_unlimited_power).

OK.  I started out wanting to replicate the lua interface.  But I have no strong opinions.  Whatever it takes to get the feature is fine with me :-)

src/main.c
bmork: > Your reviewer is an evil old-school maggot who likes to keep lines capped at 80 characters.

I usually try to match the code I modify, so I did attempt to make this line 80 chars.  But all the alternatives I tried looked awful so I gave up.

I didn't try the element pointer you propose below though.  That one looks good to me.

>     1. Do we want to power-on explicitly disabled ports? They have not been configured. Trying to turn them on has undefined results.

Good point.  I guess this should be restricted to ports being enabled in the config.

>     2. Isn't `strlen(name)` less than `sizeof(name)` ? Is there a value to str**n**cmp(), when both strings are guaranteed to be null-terminated?

Probably not.  Let's just drop the n and make this line even shorter.
 
>     3. Dereferencing an array multiple times on a line. You might want to have a pointer to the element, and dereference it directly.:
> 
> 
> ```
> port = &config[i];
> 
> if (!port->enabled || strcmp(port_name, port->name))
> ```

Yes, that does look much nicer.

src/main.c
bmork: Absolutely agree.

BUT: This was actually an attempt to copy the existing style from ubus_poe_sendframe_cb() and ubus_poe_reload_cb().  And I still think that's the correct thing to do. Mixed style is worse than bad style. And mixing refactoring with new code is really bad.

So I'd prefer it stays this way until the rest is refactored.

src/main.c
bmork: Finally, some payback for paying attention to the existing code :-)

src/main.c
bmork: Does it matter?  The code becomes more complicated with this is reversed, since we will have to finish the loop and then test again whether it succeeded or not.  We can't decide that it failed inside the loop.

Not extremely important to me, but I just like making code as simple as possible.

src/main.c
bmork: Yes. Blind cutnpaste from the ubus_poe_sendframe_cb(), without considering the effect of the shorter policy name

src/main.c
mrnuke: Scroll up... There's close to a 50-50 mix of "bad" and "preferred" style functions. You have a choice between:
1.  being consistent with the new code, submitting a good, long-lasting change, or
2.  being consistent with the old style,  submitting code that someone will have to fix later.

bmork: ok, guess you're right here as well.  Really annoying .-)  Will fix

src/main.c
mrnuke: It's not critical. Either form will work for now. This is up to you.

> We can't decide that it failed inside the loop.

But of course!
```
	for (i = 0; i < config.port_count; i++) {
		port = &config.ports[i];
		if (port->enable && strcmp(port_name, port->name)
			break;

		port = NULL;
	}
	
	if (!port)
		return UBUS_STATUS_NOT_FOUND;
```

src/main.c
mrnuke: > OK. I started out wanting to replicate the lua interface. But I have no strong opinions. Whatever it takes to get the feature is fine with me :-)

Think about it from a different point of view. The lua interface predates the contributor whose name is on the majority of commits -- He never used it, and never cared for it. To him, the "lua interface" is meaningless.

Instead, compare to the current state of realtek-poe. **Why** is the change relevant, and **what is the most logical** way to achieve it.

bmork: Sorry, I have no idea why this should be made any different.  You may say what you want about the lua implementation, but I have been an occasional user of it for the lack of anything better. Personally I care about the **feature.**  Someone who cares about interface design can do that.

bmork:
I've updated the patch with all the requested changes I know how to do.

I really don't know how/where to document the ubus interface.  AFAIK, this is how all our ubus interfaces are documented:
```
root@gs1900-10hp-f:~# ubus -v list poe
'poe' @3ca0070f
        "info":{}
        "reload":{}
        "sendframe":{"frame":"String"}
        "manage":{"port":"String","enable":"Boolean"}
```
I arbitrarily renamed the call to  "manage" in an attempt to describe the action, as requested. I believe "enable" would limit this interface unnecessarily in case we want to modify other per-port settings.  Therefore "manage", which should uniquely define what this does along with the "enable" paramater.  If there are use cases, then we can add more parameters to the same call.

src/main.c
```diff
@@ -880,10 +880,48 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
 
+static const struct blobmsg_policy ubus_poe_manage_policy[] = {
+	{ "port", BLOBMSG_TYPE_STRING },
+	{ "enable", BLOBMSG_TYPE_BOOL },
+};
+
+static int ubus_poe_manage_cb(struct ubus_context *ctx,
+			      struct ubus_object *obj,
+			      struct ubus_request_data *req,
+			      const char *method, struct blob_attr *msg)
```
mrnuke: I think you can just about fit (ctx, obj), and (req, method) on the same line.

(optional) If you don't use an argument, feel free to leave the name out. For example:
```
static int ubus_poe_manage_cb(struct ubus_context *, struct ubus_object *,
			      struct ubus_request_data *, const char *,
			      struct blob_attr *msg)

```

src/main.c
```diff
@@ -880,10 +880,48 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+		return UBUS_STATUS_INVALID_ARGUMENT;
+
+	port_name = blobmsg_get_string(tb[0]);
+	for (i = 0; i < config.port_count; i++) {
+		port = &config.ports[i];
+		if (!port->valid || strcmp(port_name, port->name))
+			continue;
+		if (!port->enable) {
+			ULOG_NOTE("Cannot manage disabled port '%s'\n", port_name);
+			return UBUS_STATUS_INVALID_ARGUMENT;
```
mrnuke: Sounds like a noble effort, but the command fails with `Command failed: Invalid argument`, and the ubus caller doesn't see the NOTE message. So why bother? Replace `!port->valid` with `!port->enable` in the first predicate, and drop the second if statement altogether. 

Yes, it's safe to assume `!valid` implies `!enable`.

mrnuke (changes requested):
Tested-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>

src/main.c
bmork: Well, the main point wasn't the log message, but making a distinction between "port does not exist" and "port cannot be managed".  For example with "lan2" disabled in config, "lan3" disabled by ubus command, and all other ports enabled:
```
root@gs1900-10hp-f:~# ubus  call poe info
{
        "firmware": "v22.4",
        "mcu": "ST Micro ST32F100 Microcontroller",
        "budget": 77.000000,
        "consumption": 15.000000,
        "ports": {
                "lan1": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan2": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Disabled"
                },
                "lan3": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Disabled"
                },
                "lan4": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan5": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan6": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Delivering power",
                        "consumption": 4.200000
                },
                "lan7": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Delivering power",
                        "consumption": 4.300000
                },
                "lan8": {
                        "priority": 0,
                        "mode": "PoE+",
                        "status": "Delivering power",
                        "consumption": 6.800000
                }
        }
}
root@gs1900-10hp-f:~# ubus  call poe manage '{"port":"lan2","enable":true}'
Command failed: Invalid argument
root@gs1900-10hp-f:~# ubus  call poe manage '{"port":"lan42","enable":true}'
Command failed: Not found
root@gs1900-10hp-f:~# ubus  call poe manage '{"port":"lan3","enable":true}'
```
Of course, if you insist, we could just return UBUS_STATUS_INVALID_ARGUMENT for all errors.  Then I agree that the "valid" test can be replaced by the "enable" test.

bmork:
OK, I ended up agreeing that there is little value in the distinction between "valid" and "enable".  Removing it makes the code much more compact.

Made one additional unrequested change: Returning the result of poe_cmd_port_enable(), similar to what the "sendframe" method does.  Not 100% sure about this since it means we return 0 instead of UBUS_STATUS_OK, but I consider consistent behaviour more important.

src/main.c
```diff
@@ -880,10 +880,40 @@ ubus_poe_reload_cb(struct ubus_context *ctx, struct ubus_object *obj,
...
+	if (!tb[0] || !tb[1])
+		return UBUS_STATUS_INVALID_ARGUMENT;
+
+	port_name = blobmsg_get_string(tb[0]);
+	for (i = 0; i < config.port_count; i++) {
+		port = &config.ports[i];
+		if (port->enable && !strcmp(port_name, port->name))
+			return poe_cmd_port_enable(i, blobmsg_get_bool(tb[1]));
+	}
+	return UBUS_STATUS_INVALID_ARGUMENT;
```
mrnuke, line 909: I sort of preferred the original NOT_FOUND return code.

mrnuke (approved):
Good enough for government work!

Hurricos:
If @mrnuke says it's good, it must be good. I have left my GS1900-24HPv1 unplugged after constructing an external serial connector for it, but once it's plugged in again I will test, Tested-by and merge.

Hurricos:
I was about to test this, then I found [this bug](https://github.com/openwrt/openwrt/issues/10530), and then I had a family emergency and needed to rush home.

Anyways. I don't think at this point I need to hold this up. Merging.

PR #26, opened 2022-08-28
[WIP] Makefile: Use AUTORELEASE for PKG_RELEASE version

mrnuke:
Let's see if this improves the package versioning.


mrnuke:
I was hoping that the $(AUTORELEASE) tag would describe the tag and number of commits since tag (similar to git describe). It instead describes the number of commits since root. This can get extremely confusing as builds from different branches may have the same package version, giving a false sense of "I'm on the right version."

For the time being, closing this.

PR #35, opened 2024-04-25
Add support for RTL82328 dialect

### LINKED ISSUE #25: Investigate feasibility of supporting RTL8238B-based protocol
### Introduction

The first PoE solution for RTL83xx switches was based on the BCM59111 family of chips. A later solution was implemented by Realtek, using the RTL8238B. realtek-poe speaks the dialect used on the former. IT does not appear to be compatible with the latter.

**Instigate the feasibility of implementing the realtek dialect in realtek-poe.**

###  Nomenclature

- **broadcom dialect** - for BCM59111 family, 19200n8
- **realtek dialect** - for RTL8238B family, 115200n8

#### Frame format

Both dialects implement a 12-byte frame format, originally documented by @svanheule [1] :
```
[ cmd_id, seq_num, {data-[0-8]}, checksum ]
```
#### Command IDs

The **realtek dialect** appears to implement completely new command IDs, simply using IDs not implemented in the **briadcom dialect**. Some commands map between dialects, albeit with a different ID

<TBD>

##### Example of an equivalent command:

Broadcom dialect:
```
TX -> 20 __ ff ff ff ff ff ff ff ff ff __
RX <- 20 __ 00 08 00 e1 11 16 00 03 0a __
```
Realtek dialect:
```
TX -> 40 __ ff ff ff ff ff ff ff ff ff __
RX <- 40 __ ff 08 01 01 38 10 01 04 11 __
```

#### Command Equivalency tables
 * **Setup commands:**
 * 
| ID | Number of occurences | Broadcom equivalent ID | Meaning |
|----|----------------------|------------------------|---------|
| 00 |                    1 | 06 | global enable
| 01 |                    8 | 00 | set port enable
| 02 |                    1 | 09 | system reset
| 03 |                      | 03 | port reset
| 04 |                    1 | 18 | set global power budget
| 05 |                      |    | port map enable (DANGEROUS)
| 06 |                      |    | port map set (DANGEROUS)
| 08 |                    8 | N/A | set port function mode or auto_powerup
| 09 |                      | 10 | set port detection mode
| 0c |                    8 | 1c | set port powerup mode
| 0f |                      | 13 | set port disconnect type
| 10 |                    1 | 17 | set power management mode
| 11 |                    1 | |  set power method extended (?)
| 12 |                    8 | 15 | set power limit type
| 13 |                    8 | 16 | set port power budget
| 14 |                    8 |   | set port power budget (0.2W steps)
| 15 |                    8 | 1a | set port priority

 * **Polling commands:**
 
| ID | Command meaning | Broadcom equivalent ID |
|----|-----------------|------------------------|
| 40 |                    2 |                 20 (?) | get system info
| 41 | get power statistics (?)  | 23 (1) |
| 42 | get meaning of the universe  and get port status | 21 |
| 43 | get port group short status  |    |
| 44 | get port measurements  | 30 |
| 45 | get port counters | 22 |
| 47 | get reset reason (?) | |
| 48 | get port config | 25 |
| 49 | get extended port config | 26 |
| 4a | get extended device config | 2b |
| 4b | get power management mode | 27 |
| 50 | get port overview | 2a (2) |

(1) The`consumed_power` is incorrect, and shows a sum of allocated power for all ports, not actual overall power consumption.
(2) The `get port overview` command only appears to work on Engenius switches. This is true for both broadcom and realtek dialects.

Command `50` (`2a` in BCM) appears to be specific to eother EnGenius or 8-port switches. Other switches do not support it and return error 0x61.

### References

[1] https://svanheule.net/switches/software/broadcom_poe_control_protocol#serial_control_protocol
[2] https://svanheule.net/switches/gpl_source_drops

mrnuke:
Add initial support for the PoE dialect used by RTL8238B PSE chips.

The first attempt at an API consisted of a struct of function pointers
representing a set of actions. A dialect would have to implement its
own set of handlers by populating these pointers. Common code could be
placed in shared functions and reused across dialects. This approach
did not allow the PoE reply handlers to be shared. Each dialect needed
to maintain a mapping of command IDs to reply_handler function.

The second attempt at an API defined a new set of actions as an enum,
and expected each dialect to provide a mapping from realtek-poe
actions to command IDs. This works well because the command IDs are
the main difference between the realtek and broadcom protocols. The
majority of existing code can be reused. The mapping can be reversed
in order to translate a command id to a generic packet handler. This
solves the per-dialect duplication of the reply handler.

However, the dialects do not map 1:1. The structure of some commands
in the realtek dialect is noticeably different from equivalent
commands in the broadcom dialect.

This creates the third and current mapping API. Use the mapping API as
the main way to differentiate dialects. For incompatible commands,
introduce introduce a structure of function pointers to allow per-
dialect customizations. This approach, hybrid as it may be, solves
the issues encountered with the previous two attempts.

```
struct dialect_ops {
	int (*init_async)(struct mcu *mcu, const struct config *cfg);
	int (*poll_async)(struct mcu *mcu, const struct config *cfg);
	int (*reset)(struct mcu *mcu);
	int (*handle_reply)(struct mcu_state *mcu, uint8_t *reply, size_t len);
};

struct dialect_map {
	const struct dialect_map_entry *entries;
	size_t len;
};

struct dialect_desc {
	const struct dialect_ops *ops;
	const struct dialect_map *map;
};
```

Many commands are very similar between the the two dialects. The command IDs differ. The approach is to have common code implementing the dialects, and have a mapping bewteen command and command_id of the dialect:

```
enum poe_cmd {
	POE_CMD_NONE = 0,
	POE_CMD_SET_XYZ,
	...
	POE_CMD_SET_YZW,
	...
};

poe_cmd_set_xyz(struct poe_dialect *dialect, ...)
{
	uint8_t cmd = dialect_map(dialect, POE_CMD_SET_XYZ);

	poe_cmd_queue(cmd, ...);
}

const struct dialect_map __map = {
	[POE_CMD_SET_XYZ] = { 0x42, CMD_4_PORT, CMD_ALL_PORT },
	...
};
```

The flags may seem obscure. Most command dealing with a port number can take more than one (port, value) argument pair. For example `01 <val>02 <val> 03 <val> ff ff`, up to four pairs per command, with `ff` meaning "nothing to see here. This is the use case for the 4port command function:

```
static int poet_cmd_4_port(uint8_t cmd_id, uint8_t port[4], uint8_t data[4])
{
	uint8_t cmd[] = { cmd_id, 0x00, port[0], data[0], port[1], data[1],
		port[2], data[2], port[3], data[3] };

		return poe_cmd_queue(cmd, sizeof(cmd));
}
```

Other commands can use `7f` as the port ID to configure all ports with the same value: `7f <val>`. The flags in `dialect` map would indicate which commands can use such optimizations.

I am hoping that restructuring the code in this manner would reduce duplication, and actually make a PR that makes some sense.


Neustradamus:
To follow this PR :)

plappermaul:
Just for completeness ...

https://github.com/plappermaul/realtek-doc/blob/main/RTL8238B%20Datasheet.pdf

https://github.com/plappermaul/realtek-doc/blob/main/RTL8238B%20Host%20Command%20Guide.pdf

mrnuke:
@Hurricos , I am ready for you to begin your review. The main blocker has been my wish to implement dialect auto-detection. However, it seems that that is a classic case of letting perfect be the enemy of good. I don't want to delay this any longer, as the realtek dialect code otherwise works okay.

There are a few changes I want to make to the data structures, but I feel that the mapping API and split between ops and raw mapping is a reliable solution at this point.

src/main.c
```diff
@@ -358,6 +358,17 @@ static int poe_cmd_port_disconnect_type(struct mcu *mcu, uint8_t port,
 	return mcu_queue_cmd(mcu, cmd, sizeof(cmd));
 }
 
+/* no BCM equivalent - Set "auto_powerup" parameter
+ *	Has no BCM, ID, but needs to be set correctly. This is one of the main
+ *	reasons RTL8238B ports get stuck on "Requesting power" status, and
+ *	never transition to "Delivering power".
+ */
+static int poe_cmd_mystery_parameter(struct mcu *mcu, uint8_t ports[4],
```
plappermaul, line 366: Thank you for enhancing the realtek-poe package. Why would we call this a mystery parameter if we have all documentation at hand? https://github.com/plappermaul/realtek-doc/blob/main/datasheets/RTL8238B%20Host%20Command%20Guide.pdf

mrnuke: Because the documentation wasn't available when I wrote the code, and i like to leave easter eggs here and there.

aparcar: I'm not entirely sure what's the best way forward here but I just got myself a 24EP and really need proper PoE support, could we solve this and merge this please :)?

plappermaul: I have no issues with this. This was just a reminder that we have mor knowledge now. Merging this PR is up to @Hurricos and @mrnuke

Hurricos:
Sorry for the lack of warning.

Sounds like there is consensus to merge here?

Hurricos:
Hmm, I'm not going to remember to merge this otherwise.

Chaos! It's merged! Hahahahahaha.

What I don't know off the top of my head is what's needed upstream. I believe we need a new package bump in openwrt/packages. @mrnuke or @aparcar or @plappermaul - could you spin up a PR there?

Klaas-:
@Hurricos I guess you need a tagged version that includes this change and then we can do a PR towards openwrt packages like https://github.com/openwrt/packages/pull/26854/files

mrnuke:
I pushed the v1.3 tag. I think the rest is as simple as changing the `PKG_VERSION` in the [packages feed](https://github.com/openwrt/packages/blob/7be94fac3708ba2ae15abcb2a92435a57b05dd37/utils/realtek-poe/Makefile#L6C1-L6C12).

hailfinger:
Something like this? https://github.com/openwrt/packages/pull/27386

PR #36, opened 2024-05-05
Eliminate global variables and clean up coding style

### LINKED ISSUE #11: Remove bash-style function declarations from C code

mrnuke:
The original realtek-poe contained a tonne of global variables. I never liked that approach, but it also didn't really get in the way. Thus there was no immediate need to invest a large effort to refactor the passing of arguments.

However, this really started to bite us in the ass when looking at adding another dialect in #25 . It makes splitting code into different files difficult, and causes a bunch of maintainability  issues that  I don't list here.

So let's improve the situation, and use proper contexts throughout realtek-poe.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -45,6 +45,11 @@ struct cmd {
 	unsigned int num_retries;
 };
 
+struct poe_ctx {
+	struct ubus_auto_conn conn;
+	struct uloop_timeout state_timeout;
+};
+
 static struct ustream_fd stream;
 static LIST_HEAD(cmd_pending);
 static unsigned char cmd_seq;
@@ -56,6 +61,12 @@ static struct config config = {
 	.budget_guard = 7,
 	.pse_id_set_budget_mask = 0x01,
 };
+
+static inline struct poe_ctx *ubus_to_poe_ctx(struct ubus_context *u)
+{
+	struct ubus_auto_conn *c = container_of(u, struct ubus_auto_conn, ctx);
+	return container_of(c, struct poe_ctx, conn);
+}
 
 static void load_port_config(struct uci_context *uci, struct uci_section *s)
 {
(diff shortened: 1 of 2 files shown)


Hurricos:
ea32075ae41874043eeed14156755a4c0f114582 created a conflict. It's dumb and it's simple.

```diff
diff --cc src/main.c                                                                   
index 7a1fa7a,eed5c6c..0000000                                                         
--- a/src/main.c                                                                       
+++ b/src/main.c                                                                                                                                                              
@@@ -25,34 -24,7 +24,11 @@@ typedef int (*poe_reply_handler)(unsign                                                                                                           
                                                                                       
  /* Careful with this; Only works for set_detection/disconnect_type commands. */                                                                                             
  #define PORT_ID_ALL   0x7f                                                                                                                                                  
++<<<<<<< HEAD                                                                                                                                                                
 +#define MAX_PORT      48                                                             
++=======                                                                              
++>>>>>>> 9f5b7c0 (realtek-poe: move protocol-agnostic code to common header)         
```

I hate merge commits. They make me sad. So, I rebased mrnuke-tek-poe-cleanups onto realtek-poe, then force-pushed.

Hurricos:
Merged into realtek-poe.

PR #39, opened 2024-07-18
Increase MAX_PORT to 48 for larger devices

Hurricos:
This accommodates the HPE 1920-48G-PoE (JG928A) 48-port PoE switch. Increasing MAX_PORT only requires a bit more stack memory for all platforms -- we don't for example do any kind of repeated discovery iterating over this variable, so there's no real downside here.

The upside is, of course, supporting these platforms.


howels:
Please merge and release - key to enabling PoE support on larger switches.

Issue #41, opened 2024-10-19
replace static PoE config with generator based on board.json contents


howels:
Currently the static config is copied to /etc/config/poe and the PoE definitions in /etc/board.json are not processed.

Config file should be generated dynamically e.g. when running `ucidef_set_poe` such as shown in https://github.com/openwrt/openwrt/blob/main/target/linux/realtek/base-files/etc/board.d/02_network

```
ucidef_set_poe 370 "lan8 lan7 lan6 lan5 lan4 lan3 lan2 lan1 lan16 lan15 lan14 lan13 lan12 lan11 lan10 lan9 lan24 lan23
			lan22 lan21 lan20 lan19 lan18 lan17 lan32 lan31 lan30 lan29 lan28 lan27 lan26 lan25 lan40 lan39 lan38 lan37
			lan36 lan35 lan34 lan33 lan48 lan47 lan46 lan45 lan44 lan43 lan42 lan41"
```

mrnuke:
It's not the job or realtek-poe to parse the JSON config. See https://github.com/Hurricos/realtek-poe/pull/31#issuecomment-1854861920

> I think the contract of realtek-poe is to take a config file and control the hardware. We speak UCI config on one end, and the PoE protocol on the other.
>
> I don't think it should be the job of realtek-poe to generate the config files. The question then arises: If we speak UCI. and JSON, why not take the JSON config in the first place?

PR #42, opened 2024-12-13
realtek-poe/openwrt: Generate config from board.json

Hurricos:
Styled after work provided by @svanheule, then reiterated by @howels.

Before this commit, realtek-poe maintainers prioritized keeping realtek-poe simple; it was "[a tool to] take a config file and control the hardware ... [which speaks] UCI config on one end, and PoE protocol on the other"[^1]. We declined to support generating that config file previously.

OpenWrt maintainer @blogic firmly disagreed[^2], because OpenWrt users were finding their devices starting with non-functional PoE (which is a violation of the principle of least surprise).

It's been about a year since a solution to this problem was first provided[^3].

Change stance. Now, begin generating /etc/config/poe from board.json under OpenWrt (our only supported platform). It's in the community's best interest to at least provide a stop-gap until either:

- other software actually uses realtek-poe outside of OpenWrt, or
- another OpenWrt package uses /etc/config/poe to configure these PSEs

at which point I'll revisit and do the integration work.

A companion commit will be required to openwrt/packages.git.

Signed-off-by: Martin Kennedy <hurricos@gmail.com>

[^1]: https://github.com/Hurricos/realtek-poe/issues/41#issuecomment-2426670039

[^2]: https://github.com/openwrt/openwrt/pull/16838#issuecomment-2542114739

[^3]: https://github.com/Hurricos/realtek-poe/pull/31

Replaces: #31

### DIFF
--- a/files/etc/uci-defaults/30-poe
+++ b/files/etc/uci-defaults/30-poe
@@ -1,0 +1,42 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+[ -e /etc/config/poe ] && exit 0
+
+. /lib/functions.sh
+. /usr/share/libubox/jshn.sh
+
+CFG=/etc/board.json
+
+json_init
+json_load_file "${CFG}"
+
+json_is_a poe object || exit 0
+
+umask 055
+touch /etc/config/poe
+
+json_select poe
+	json_get_vars budget
+
+	uci add poe global
+	uci set poe.@global[-1].budget="$budget"
+
+	if json_is_a ports array; then
+		json_get_values ports ports
+		id='1'
+
+		for port in $ports; do
+			uci -q batch <<-EOF
+				add poe port
+				set poe.@port[-1].name='$port'
+				set poe.@port[-1].id='$id'
+				set poe.@port[-1].enable='1'
+			EOF
+			let id=id+1
+		done
+	fi
+json_select ..
+
+uci commit
+
+exit 0


Hurricos:
@mrnuke, @svanheule: on a scale from 0 to incredible psychic pain, how much hurt does this PR cause you?

My take: let's get something out-the-door, to staunch the bleeding in time for the OpenWrt 24 final builds.

howels (approved):

howels:
A pragmatic solution, thanks.

mrnuke:
I don't think asking for an opinion from the main contributors and then only giving them 50 minutes to respond before merging something is reasonable or collaborative.

As a separate project, realtek-poe is upstream to OpenWRT. It's also a special relationship because they are realtek-poe's only downstream user. Be it as it may, while downstream projects may request and influence upstreams, they don't get to dictate what upstream does. I disagree with @blogic, and I do not find any objective argument in his request that realtek-poe takes on the responsibility of parsing an arbitrary JSON format.

I am of the mindset of do one thing and do it well. And by pushing this commit, this principle has been violated.

Hurricos:
> I don't think asking for an opinion from the main contributors and then only giving them 50 minutes to respond before merging something is reasonable or collaborative.

Partially fixed with 89686c9d625ab515868a8152dc5010e338435784.

svanheule:
Placing the uci-defaults script in the packages feed seems like a decent intermediate solution to me. That way the config format stays specific to poe-realtek, until we get a better understanding of common requirements.

PR #44, opened 2025-01-11
realtek-poe: info: add voltage, current and temperature

grobian:
Expose voltage, current and temperature readings from each port.

On my HPE-1920 JG926A the values are returned and contain seemingly useful data:

       "lan3": {
        "priority": 2,
        "mode": "PoE+",
        "status": "Delivering power",
        "voltage": 52.397852,
        "current": 0.091000,
        "temperature": 20.000000,
        "consumption": 4.700000
       },
...
       "lan13": {
        "priority": 2,
        "mode": "PoE+",
        "status": "Delivering power",
        "voltage": 52.462301,
        "current": 0.092000,
        "temperature": 21.250000,
        "consumption": 4.800000
       },
...
       "lan23": {
        "priority": 2,
        "mode": "PoE+",
        "status": "Searching",
        "temperature": 11.250000
       },

The temperature seems to be always available, and per block of 8 (controller?) the same value.  The values make sense (in Celcius) and I guess could be used to detect overheating and taking actions (up the fans, disable power output).

The voltage and current are just nice to have, and since it's available for free anyway, why not.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -498,6 +498,9 @@ static int poe_reply_port_power_stats(struct mcu_state *state, uint8_t *reply)
 {
 	int port_idx = reply[2];
 
+	state->ports[port_idx].mvolt = read16_be(reply + 3) * 64.45;
+	state->ports[port_idx].mampere = read16_be(reply + 5);
+	state->ports[port_idx].tempc = (220 - read16_be(reply + 7)) * 1.25;
 	state->ports[port_idx].watt = read16_be(reply + 9) * 0.1;
 	return 0;
 }
@@ -806,6 +809,12 @@ static int ubus_poe_info_cb(struct ubus_context *ctx, struct ubus_object *obj,
 			blobmsg_add_string(b, "status", state->ports[i].status);
 		else
 			blobmsg_add_string(b, "status", "unknown");
+		if (state->ports[i].mvolt)
+			blobmsg_add_double(b, "voltage", state->ports[i].mvolt / 1000.0);
+		if (state->ports[i].mampere)
+			blobmsg_add_double(b, "current", state->ports[i].mampere / 1000.0);
+		if (state->ports[i].tempc)
+			blobmsg_add_double(b, "temperature", state->ports[i].tempc);
 		if (state->ports[i].watt)
 			blobmsg_add_double(b, "consumption", state->ports[i].watt);
 
--- a/src/tek-poe.h
+++ b/src/tek-poe.h
@@ -15,6 +15,9 @@
 
 struct port_state {
 	const char *status;
+	float mvolt;
+	float mampere;
+	float tempc;
 	float watt;
 	const char *poe_mode;
 };


grobian:
I also tested this on a ZyXEL GS1900-8HP, it also always returns temperature readings, they seem to differ per port (or: 1 and 8 report a different temp than the rest, but the temp is still in line with what could be correct).  The voltage and current readings work fine for a connected port there too.

src/tek-poe.h
```diff
@@ -15,6 +15,9 @@
 
 struct port_state {
 	const char *status;
+	float mvolt;
+	float mampere;
```
mrnuke, line 19: Just use SI units here  -- `voltage` and `current` for variable names with units [V], and [A] respectively. For someone without scientific training, it may not be obvious if it's micro, milli, or mega.

src/tek-poe.h
```diff
@@ -15,6 +15,9 @@
...
+	float tempc;
```
mrnuke, line 20: `temp_celsius` if you intend to keep the temperature.

src/main.c
```diff
@@ -806,6 +809,12 @@ static int ubus_poe_info_cb(struct ubus_context *ctx, struct ubus_object *obj,
 			blobmsg_add_string(b, "status", state->ports[i].status);
 		else
 			blobmsg_add_string(b, "status", "unknown");
+		if (state->ports[i].mvolt)
+			blobmsg_add_double(b, "voltage", state->ports[i].mvolt / 1000.0);
+		if (state->ports[i].mampere)
+			blobmsg_add_double(b, "current", state->ports[i].mampere / 1000.0);
+		if (state->ports[i].tempc)
+			blobmsg_add_double(b, "temperature", state->ports[i].tempc);
```
mrnuke, line 817: Only report the temperature if it's not 0 C? This doesn't look right

mrnuke (changes requested):
Thank you. I had completely forgotten that the protocol gives us voltage and current.

Why is the temperature measurement useful? The sensors are local to the PoE controller, so what can they tell us about the PoE sinks? There is a lot of minor information that we have but don't report for this reason.

src/main.c
```diff
@@ -806,6 +809,12 @@ static int ubus_poe_info_cb(struct ubus_context *ctx, struct ubus_object *obj,
 			blobmsg_add_string(b, "status", state->ports[i].status);
 		else
 			blobmsg_add_string(b, "status", "unknown");
+		if (state->ports[i].mvolt)
+			blobmsg_add_double(b, "voltage", state->ports[i].mvolt / 1000.0);
+		if (state->ports[i].mampere)
+			blobmsg_add_double(b, "current", state->ports[i].mampere / 1000.0);
```
mrnuke, line 815: My funny commit about reporting superconductor-like behavior (zero voltage, non-zero current) got lost by github. Short story long, report current and voltage together, under the same predicate:

```
		if (state->ports[i].voltage > 0.0) {
			blobmsg_add_double(b, "voltage", state->ports[i].voltage);
			blobmsg_add_double(b, "current", state->ports[i].current);
		}
```

grobian:
> Why is the temperature measurement useful? The sensors are local to the PoE controller, so what can they tell us about the PoE sinks? There is a lot of minor information that we have but don't report for this reason.

Well, basically because the values are returned per port, and under certain conditions differ per port, as I witnessed with GS1900-8HP.

src/tek-poe.h
grobian: Well, I chose this because the protocol returns mV and mA, and I wanted to stay close to the unit the protocol documents.
Nevertheless I can normalise it to volts and amperes too.

src/tek-poe.h
grobian: that's what it is according to the protocol, yes

src/main.c
grobian: got it, I'll see if I can map an absent value to nan.

src/main.c
grobian: I don't understand this.  If the protocol reports 0 voltage and still a current, then it must mean there is a rounding problem, or the implementation doesn't return the voltage, apparently.
I also noticed (you can calculate it off my example trace) that the watt value is not the same as the calculation of volt * ampere (as it should).

So I believe just returning the values, not trying to interpret/match/understand them is better here, because YMMV.

mrnuke: How logical is it to output < 63 mV on a 48V bus, and have a measurable current? That is considered a short circuit in 802.3at/bt terms, and indicated as a fault in the port status. So not, the suggested use case does not make sense.

In a perfect world, one would report V-A if and only if the state is "Delivering power". That's impractical because the status and the measurements arrive to us at different times. That could lead to race conditions such as "delivering power 0.0W". So the next best heuristic is `port->voltage > 0.0`.

mrnuke:
> Well, basically because the values are returned per port, and under certain conditions differ per port, as I witnessed with GS1900-8HP.

There are a lot of parameters that we know, that we don't report in the `info` data. For example, see
https://github.com/Hurricos/realtek-poe/blob/12fec5d509c5b83cdafcf0621f5a0a9ee00bfb32/src/main.c#L455

We don't report everything in `info` because it's beyond the scope of reporting the status of the power delivery. So why is the temperature relevant to the port status message instead of a debugging or troubleshooting message?

grobian:
Ok, it seems to me you're not interested in this, that's fine.  Just close this PR.  It was an experiment for me to see how to get changes running on the real hardware.  My real goal was to enable PoE LEDs, which I'm quite far progressed with.  Judging by your comments I don't think you'll like that either so I'll just keep them in my fork and enjoy the information and functionality.

mrnuke:
We (@Hurricos and I) are interested. A "take it or leave it" approach won't help anyone. Here's what I propose:

- limit this PR to V-A info. Report either both V and A or none. 
- If you want the temperature, we can add it to the "debug" packet

grobian:
The temperature allows you to make decisions per PSE about whether you want to start dropping delivering power to ports in terms of their priority.  That combined with the power draw per PSE (currently not retrieved) should probably do it.  Having 370W total does not mean it can deliver that through a single port.  Having half of it on a single PSE is likely also a condition that might not be healthy for the device.

evs93:
It would be nice to be able to integrate temperature into a fan control algorithm for JG922A, JG925A,JG926A,JG928A. Looking forward to these changes. 

What is the temperature being reported? 11 degrees or 21 degrees sounds low to me unless it's an ambient temperature?

Regarding voltage/current. I can only think about under/over voltage at the PSE? should be 50V-57V for 802.3at? Not sure if this is just a failure scenario or whether there are other targets that have external power supply only? (Not sure how EPS works on JG926A/JG928A).

Current can be found from voltage and power anyway as we're at DC? Anyway power consumption appears to be truncated to one decimal place? i.e. multiplying the reported voltage/current on the JG926A example gives 4.77W?

grobian:
> What is the temperature being reported? 11 degrees or 21 degrees sounds low to me unless it's an ambient temperature?

It makes sense.  It's cold outside here currently, and the basement the switch is in, is about 10 degrees Celcius.  The idle ports being 11 makes sense from that point.  The 21 for ones that are used too, it's a narrow case, and I set the fans to 5000RPMs.  While my readings appear per block for the HPEs, they should be per port, as per specification.  A ZyXEL 8HP shows that behaviour.

> Regarding voltage/current. I can only think about under/over voltage at the PSE? should be 50V-57V for 802.3at? Not sure if this is just a failure scenario or whether there are other targets that have external power supply only? (Not sure how EPS works on JG926A/JG928A).

After reading the BCM feature spec, it seems voltage and current are the only "real" things, because power isn't mentioned, probably it's calculated by the PSE or something.  The thing gets interesting when they pair two PSEs, the 380W variants apparently can connect to a backup power and double whatever it can deliver.

> Current can be found from voltage and power anyway as we're at DC? Anyway power consumption appears to be truncated to one decimal place? i.e. multiplying the reported voltage/current on the JG926A example gives 4.77W?

Yeah, I figured that too, but the value is off by a little.  This can be due to a number of things, perhaps rounding.

src/tek-poe.h
svanheule: Alternatively, you could use `voltage_mv` and `current_ma`.

src/tek-poe.h
svanheule: Certain parts of the world might like a redundant `temp_fahrenheit`  too.

svanheule:
> The temperature seems to be always available, and per block of 8 (controller?) the same value.

That would make sense if the temperature sensor is baked into the BCM PSE chips, so there is only one per eight outputs.

The details currently provided by `info` could be interpreted as reflecting more the PSE-PD link state. Voltage, current, and temperature are local properties at the PSE. Maybe it could be useful to have a separate `pse_info` call that reports this info without requiring a full debug dump?

Not all PSE controllers would report the same details, so a hypothetical `pse_info` would have most if not all properties optional. At the same time `info` could remain as it is now, with little if no optional fields.

src/tek-poe.h
grobian: I like that suggestion

mrnuke:
> Not all PSE controllers would report the same details, so a hypothetical `pse_info` would have most if not all properties optional. At the same time `info` could remain as it is now, with little if no optional fields.

That's exactly what I was asking for in the first place. I went ahead and pulled the debug info from my 8238 branch in #45 .

src/tek-poe.h
grobian: Yeah, but that would be more something for the rendering stage, I think.  Could report double, or use config to decide.

src/tek-poe.h
mrnuke: No. Use SI units, no prefix.

grobian:
obsoleted by https://github.com/Hurricos/realtek-poe/pull/45 .

mrnuke:
@grobian , #45 does not add the voltage, current, or temperature measurements. This PR is still very much relevant. It also allows you the chance to add them, so that you may get attribution for the idea.

grobian:
Look, it was a test on my end, 5 lines of code, and you had remarks about each and every one of them.  If you really think it's useful to have, just add it yourself the way you think is right.  I'm obviously on an entirely different plane for tihis code repository.

src/tek-poe.h
Hurricos: Definitely use SI units instead of mA/mV -- remember, our choices here will echo everywhere; as a daemon, realtek-poe's opinions are the standard in the OSS PoE control world for all of these PSEs. If these items are subsequently pulled into the Prometheus exporter for realtek-poe, imagine the displeasure metrics-writers might have having to multiply the wattage they get as a product by 1E6.

grobian: Don't see what this has to do with it.  This struct stores a value reading.  The way in which we present the value is completely unrelated, and handled in the formatting.  Internal value following the hardware/remote unit vs exposed value which is conforming standards, previous behaviours, etc.

Hurricos:
Hey @grobian,

As far as I can tell, the PR is quite close as-is. I would ask you reopen it.

Don't forget that the function of code-review is to make sure every aspect of the code can stand on its own (and that other contributors have space to adapt code patterns to their own implementations, as with the rtl8238 branch); and that we are not making arbitrary (read: unstable) changes -- this is especially important seeing as realtek-poe acts as a standards-setter for PSE in OpenWrt.

Hurricos:
> Yeah, I figured that too, but the value is off by a little. This can be due to a number of things, perhaps rounding.

Just as likely that there's a race / sampling issue, yeah.

src/tek-poe.h
Hurricos: I've re-read the actual change and, yes, I see your point; *(edited) sorry to have come back at you without good talking points*.

Still, the `port_state` struct is used in a few places, and doesn't elsewhere reflect the protocol, but instead a post-processed view atop of it (e.g. `port_state.poe_mode`).

Hurricos:
> We (@Hurricos and I) are interested. A "take it or leave it" approach won't help anyone. Here's what I propose:
> 
>     * limit this PR to V-A info. Report either both V and A or none.
> 
>     * If you want the temperature, we can add it to the "debug" packet

+1. It is my understanding that adding it *somewhere* will mean it's available in our structs, and once we've got a better pattern to understand *what* the temperature is reflecting on a per-model, per-PSE model or per-MCU-FW basis, we can publish it.

Hurricos:
(I also don't like the idea of controlling thermals from a userspace program -- it should be done by hardware before by realtime software before by the Linux kernel before by userspace. Still, it sounds like ideal design philosophy took a back seat at every step until these switches dropped into our laps.)

grobian:
> +1. It is my understanding that adding it _somewhere_ will mean it's available in our structs, and once we've got a better pattern to understand _what_ the temperature is reflecting on a per-model, per-PSE model or per-MCU-FW basis, we can publish it.

... which justifies dropping these changes

It's just a reading, remarkably interesting to some, raising all kinds of questions to others.  Not worth a huge debate at all.  I rather focus the attention to the real changes I did in the other PRs.

grobian:
> (I also don't like the idea of controlling thermals from a userspace program -- it should be done by hardware before by realtime software before by the Linux kernel before by userspace. Still, it sounds like ideal design philosophy took a back seat at every step until these switches dropped into our laps.)

This is a bit of a mistaken (on my end) concept.  Exceeding the power draw should be handled by the MCU, as far as I understand.  It uses the priorities, and just drops ports.

The temperature bit came from the HPE 1920's which have controllable fans, running in a low or high state.  The vendor firmware will switch the fans to low when the load or temp is low.  Not sure how/what, but after a few minutes after boot, the fans come down.  I simulated this behaviour currently by looking at the consumed power as opposed to the allocated budget.  When its <50% I just lower the fans via a script that runs every minute via cron.  I wondered whether I could factor in the temperature per block/PSE or on some ports, and trigger on that as well, but since we're unsure about the concept temperature and I don't really know what would be a reasonable low/bound I guess for my usage the current 50% rule is sufficient.

Neustradamus:
Dear all,

What is the situation of this PR?

Hurricos:
Rereading now. Two ways to look at it:

- I believe after some not-contention-free dialogue, we realized this PR might not be useful; we identified why it might be folly to export temperature, voltage and amperage. (Whether it's because our users will try to use these pieces of data badly, etc.)
- This data **could** be useful, and the pull-request process lead to contention about particulars over which @grobian didn't want to overnitpick.

Personally, I've never seen PSE ports fail non-dramatically, and I have no idea whether supplying this data would help with it. Personally, I'd go with closing the PR, but I know @mrnuke thinks the data might be useful if we get all the units right.

Still, ... @Neustradamus though I can't see your github history, I suspect your wedge here is that you want all features merged and pushed up before a package release you have a vested interest in (good! that's open-source for you!).

I'm not yet convinced these pieces of data makes for a really useful feature.

Neustradamus:
@Hurricos: @grobian PR adds good informations.
Maybe to add an option to disable/enable informations?
And not enabled by default?

grobian:
I appreciate it, but really, it just exposes some more random readings.  It's not something that improves the functionality of realtek-poe in any way.  This is cosmetics, if there's going to be a way in which it is acceptible/right to provide the values, then this PR may be used as inspiration by anyone who would like to see those values show up at some point.

Hurricos:
> This is cosmetics, if there's going to be a way in which it is acceptible/right to provide the values, then this PR may be used as inspiration by anyone who would like to see those values show up at some point.

This is the take that makes the most sense to me: closing this PR does not prevent a motivated user-developer from deciding they need this info, finding how to pull it in this PR, and then reopening or opening a new PR.

Closing. Thanks for screwing my head on right, @grobian!

PR #46, opened 2025-01-13
realtek-poe: port_ext_config: move from timeout loop to setup

grobian:
Reduce traffic/stress (and thus cpu intensity/load) a bit by not retrieving extended port config all the time.  The only thing we currently monitor is the powerup mode, which won't change unless we set it.  So move it to setup where it will be re-run if the config changes.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -816,8 +816,10 @@ static int poe_port_setup(struct mcu* mcu, const struct config *cfg)
 
 	poet_setup(mcu, cfg->ports, cfg->port_count);
 
-	for (i = 0; i < cfg->port_count; i++)
+	for (i = 0; i < cfg->port_count; i++) {
 		poe_cmd_port_enable(mcu, i, !!cfg->ports[i].enable);
+		poe_cmd_port_ext_config(mcu, i);
+	}
 
 	return 0;
 }
@@ -854,6 +856,12 @@ static void state_timeout_cb(struct uloop_timeout *t)
 	struct mcu *mcu = &poe->mcu;
 	size_t i;
 
+	/* skip this iteration if we're still busy processing the queue */
+	if (!list_empty(&mcu->pending_cmds)) {
+		uloop_timeout_set(t, 1 * 1000);
+		return;
+	}
+
 	poe_cmd_power_stats(mcu);
 	if (poe->hardcore_hacking_mode_en)
 		poe_cmd_get_extended_config(mcu);
@@ -867,7 +875,6 @@ static void state_timeout_cb(struct uloop_timeout *t)
 			poe_cmd_port_config(mcu, i);
 		}
 
-		poe_cmd_port_ext_config(mcu, i);
 		poe_cmd_port_power_stats(mcu, i);
 	}
 


mrnuke (approved):
My main concern is that I'd like this tested on a 48-port switch. I'm not sure how the MCU will respond if flooded with commands at init time. While it should be handled gracefully with 5e72913c41e8f22091b83ef58896c0aff9a9a0bc, I'd be more confident with a real world test.

grobian:
Perhaps I should plug in my change that I made in my repos that actually makes the timeout_cb skip a run when the queue isn't empty.  I observed that realtek-poe on my 24-ports switch can really keep things busy, and skipping a cycle when the queue hasn't been drained at least allowed it to catch up quicker, instead of infinitely adding new requests.

That said, realtek-poe uses a lot of cpu compared to other deamons, and I wonder if the loop frequency/timeout shouldn't be just longer to reduce the load.  Regular monitoring intervals are around 1min, but perhaps a config option to set the cycle time would allow any usecase?

Thanks,
Let me know.

mrnuke:
The "poll every `n` seconds" heuristic is fairly arbitrary, which comes all the way from the initial commit b8dc3ce. If packets are queued at regular intervals, without time to send them all, the queue grows without bound. That would a problem. Is that what you are seeing?

A cycle time as a config option doesn't fully solve the problem if that config is not done right. Obviously we shouldn't be enqueuing a new set of probe packets if the old one isn't finished. What I'm thinking is "send a probe when the queue is empty, but no more often than `n` seconds.

> realtek-poe uses a lot of cpu compared to other deamons,

Does it use up more than 1% CPU as shown in `top` or `htop`. That sounds like a bug.

grobian:
In debug mode it rarely is "quiet"/waiting.

Here's top's output, 23.05.5 release, this switch is 1 day up:
```
  PID USER      PR  NI    VIRT    RES  %CPU  %MEM     TIME+ S COMMAND          
 1928 root      20   0    1.5m   1.0m   1.3   0.8  23:56.82 S  `- /usr/bin/rea+
```
I think that's pretty heavy almost 24 minutes of usage, while a metric forwarding proxy which does regex matching and processing on monitoring strings on the same switch "only" took 4 minutes.

grobian:
I've added the stopgap, please look at ~754

grobian:
with this patch the cpu usage is around 0.6 by the way

mrnuke:
I was thinking more along the lines of sending the new probes immediately when `(timeout && queue_empty)` . With an extra timeout, we only check for the condition once a second. But that's still a valid solution.

On a serial connection, whenever a byte is received, realtek-poe is woken up, and `poe_stream_msg_cb()` is called. With that in mind, 1% CPU usage on a 500MHz core doesn't seem that insane. One can try to tune VMIN and VTIME, and deal with blovking reads, but it will only save up to 1% CPU.

mrnuke:
@Hurricos , ping. This change LGTM

Hurricos:
> My main concern is that I'd like this tested on a 48-port switch. I'm not sure how the MCU will respond if flooded with commands at init time. While it should be handled gracefully with [5e72913](https://github.com/Hurricos/realtek-poe/commit/5e72913c41e8f22091b83ef58896c0aff9a9a0bc), I'd be more confident with a real world test.

I have a 48-port PoE switch set up remotely. Regrettably I've left it unplugged. I'll return to it in two weeks - marked on my calendar.

Are there any ways to increase the traffic on init that we should be testing? Extended priority settings, or PoE type settings?

howels:
> > My main concern is that I'd like this tested on a 48-port switch. I'm not sure how the MCU will respond if flooded with commands at init time. While it should be handled gracefully with [5e72913](https://github.com/Hurricos/realtek-poe/commit/5e72913c41e8f22091b83ef58896c0aff9a9a0bc), I'd be more confident with a real world test.
> 
> I have a 48-port PoE switch set up remotely. Regrettably I've left it unplugged. I'll return to it in two weeks - marked on my calendar.
> 
> Are there any ways to increase the traffic on init that we should be testing? Extended priority settings, or PoE type settings?

I've tested a similar fix after observing high load and memory issues on my 48 portable PoE switches, leading to reboots. I'll test this PR too, hopefully it can resolve the reboots.

howels:
Working reliably and keeping RAM usage for userspace safely under 30MB (including Luci and Prometheus node exporter) whereas the prior version would peak to 70MB and trigger reboots every 1-2 days. No issues so far.

howels (approved):
Working reliably, preventing reboots on 48 port switch.

Please review the OpenWRT packages PR @mrnuke @Hurricos https://github.com/openwrt/packages/pull/26789

mrnuke:
Merged. Thank you @grobian for the submission, and @howels for testing.

PR #47, opened 2025-01-13
realtek-poe: logging: don't log to stdio by default

grobian:
Only push log messages to stdio when debugging (-d).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -937,6 +937,8 @@ static void ubus_connect_handler(struct ubus_context *ctx)
 int main(int argc, char **argv)
 {
 	int ch;
+	int treshold = LOG_INFO;
+	int channels = ULOG_SYSLOG;
 
 	struct poe_ctx poe = {
 		.state_timeout.cb = state_timeout_cb,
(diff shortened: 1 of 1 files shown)


Neustradamus:
@Hurricos: Have you seen this @grobian PR?

Hurricos:
Hmmmmh ... I read the "don't log to stdio" as "don't log to stdout," but I'm less familiar than I'd like to be with libubox.

Now that I know what you mean (`ULOG_STDIO`), I feel like I should disagree. I know it's ugly to see piles of trash in `logread` on OpenWrt boxes, but having it deployed generally, I feel, does help us find patterns.

I mean, logging to STDIO was intentional - see PR #20.

I'm leaving this comment here as a tombstone, WITHOUT closing the PR: I don't think we should apply this PR.

mrnuke:
I ignored this PR because I didn't understand the motivation behind it. The commit message doesn't tell me **why** this should be applied. Why should we not log to stdio? How does this affect debug builds?

grobian:
As a daemon, you shouldn't write to stdio, you should just log.  My change doesn't mean changes disappear from `logread`, it just removes useless writing to stdout that's dev-nulled by the process launcher or something

mrnuke:
Don't we capture stdio with procd? Or did we do away with that?

PR #48, opened 2025-01-13
realtek-poe: leds: implement led config retrieval

grobian:
This implements an ubus callback "leds" that retrieves the status from the controller.  This is an informative action that may be used in the future for debugging (new) configurations.

Example output for ubus call poe leds:

HPE 1920-24G PoE JG926A:
```
{
        "system": {
                "sys_ok": "on",
                "in_gb": "off",
                "out_of_gb": "blink slow",
                "exceeds_ps": "blink fast",
                "out_of_gb_off_delay": 5,
                "exceeds_ps_off_delay": 5,
                "map_enable": false
        },
        "map": {
                "map_0_offset": "error",
                "lan8": 0,
                "lan7": 0,
                "lan6": 0,
                "lan5": 0,
                "lan4": 0,
                "lan3": 0,
                "lan2": 0,
                "lan1": 0,
                "map_1_offset": "error",
                "lan16": 0,
                "lan15": 0,
                "lan14": 0,
                "lan13": 0,
                "lan12": 0,
                "lan11": 0,
                "lan10": 0,
                "lan9": 0,
                "map_2_offset": "error",
                "lan24": 0,
                "lan23": 0,
                "lan22": 0,
                "lan21": 0,
                "lan20": 0,
                "lan19": 0,
                "lan18": 0,
                "lan17": 0
        },
        "port": {
                "enable": false,
                "interface": "SPI",
                "shift_order": "LSB",
                "port_led_count": 1,
                "state_off": "led off",
                "state_req": "led on",
                "state_err": "led on",
                "state_on": "led on",
                "blink_override": "blink when port status is
'requesting' or 'fault'"
        }
}
```

ZyXEL GS1900-8HP v1
```
{
        "system": {
                "sys_ok": "on",
                "in_gb": "off",
                "out_of_gb": "blink slow",
                "exceeds_ps": "blink fast",
                "out_of_gb_off_delay": 5,
                "exceeds_ps_off_delay": 5,
                "map_enable": false
        },
        "map": {
                "lan1": 0,
                "lan2": 1,
                "lan3": 2,
                "lan4": 3,
                "lan5": 4,
                "lan6": 5,
                "lan7": 6,
                "lan8": 7
        },
        "port": {
                "enable": true,
                "interface": "SPI",
                "shift_order": "LSB",
                "port_led_count": 2,
                "state_off": "led 1 off, led 2 off",
                "state_req": "led 1 off, led 2 on",
                "state_err": "led 1 off, led 2 on",
                "state_on": "led 1 on when PoE+, led 2 on when PoE",
                "blink_override": "none"
        }
}
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -502,12 +502,95 @@ static int poe_reply_port_power_stats(struct mcu_state *state, uint8_t *reply)
 	return 0;
 }
 
+/* 0x42 - Get port LED config */
+static int poe_cmd_port_led_config(struct mcu *mcu, uint8_t port)
+{
+	uint8_t cmd[] = { 0x42, 0x00 };
+
+	return mcu_queue_cmd(mcu, cmd, sizeof(cmd));
+}
+
+static int poe_reply_port_led_config(struct mcu_state *state, uint8_t *reply)
+{
+	state->portledconfig.enable         = reply[0x2] != 0;
+	state->portledconfig.interface      = reply[0x3] & 0x1;
+	state->portledconfig.shift_order    = reply[0x4] & 0x1;
+	state->portledconfig.led_count      = reply[0x5] & 0x3;
+	state->portledconfig.state_off      = reply[0x6];
+	state->portledconfig.state_req      = reply[0x7];
+	state->portledconfig.state_err      = reply[0x8];
+	state->portledconfig.state_on       = reply[0x9];
+	state->portledconfig.blink_override = reply[0xa] & 0x3;
+
+	return 0;
+}
+
+/* 0x44 - Get system LED config */
+static int poe_cmd_system_led_config(struct mcu *mcu)
+{
+	uint8_t cmd[] = { 0x44, 0x00 };
+
+	return mcu_queue_cmd(mcu, cmd, sizeof(cmd));
+}
+
+static int poe_reply_system_led_config(struct mcu_state *state, uint8_t *reply)
+{
+	state->sysledconfig.sys_ok               = reply[0x2] & 0x1;
+	state->sysledconfig.in_gb                = reply[0x3] & 0x3;
+	state->sysledconfig.out_of_gb            = reply[0x4] & 0x3;
+	state->sysledconfig.exceeds_ps           = reply[0x5] & 0x3;
+	state->sysledconfig.out_of_gb_off_delay  = reply[0x6];
+	state->sysledconfig.exceeds_ps_off_delay = reply[0x7];
+	state->sysledconfig.map_enable           = reply[0x8];
+
+	/* normalise map_enable value */
+	if (state->sysledconfig.map_enable == 0xff)
+		state->sysledconfig.map_enable = 0;
+
+	return 0;
+}
+
+/* 0x49 - Get port LED map */
+static int poe_cmd_port_led_map(struct mcu *mcu, uint8_t port)
+{
+	uint8_t cmd[] = { 0x49, 0x00, port };
+
+	/* can/should only call this once for every set of 8 */
+	if (port % 8 != 0)
+		return 1;
+
+	return mcu_queue_cmd(mcu, cmd, sizeof(cmd));
+}
+
+static int poe_reply_port_led_map(struct mcu_state *state, uint8_t *reply)
+{
+	struct port_led_map *map = &state->ledmaps[reply[0x2] / 8];
+
+	map->offset = reply[0x2];
+
+	/* unroll the loop, it's always going to be 8 and not worth the
+	 * effort to set it programmatically */
+	map->ports[0] = reply[0x3];
+	map->ports[1] = reply[0x4];
+	map->ports[2] = reply[0x5];
+	map->ports[3] = reply[0x6];
+	map->ports[4] = reply[0x7];
+	map->ports[5] = reply[0x8];
+	map->ports[6] = reply[0x9];
+	map->ports[7] = reply[0xa];
+
+	return 0;
+}
+
 static poe_reply_handler reply_handler[] = {
 	[0x20] = poe_reply_status,
 	[0x23] = poe_reply_power_stats,
 	[0x26] = poe_reply_port_ext_config,
 	[0x28] = poe_reply_4_port_status,
 	[0x30] = poe_reply_port_power_stats,
+	[0x42] = poe_reply_port_led_config,
+	[0x44] = poe_reply_system_led_config,
+	[0x49] = poe_reply_port_led_map,
 };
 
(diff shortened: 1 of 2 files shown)


Hurricos:
I'm dumb as bricks and have been a little less involved for a while, so forgive the dumb questions:

1) What's the output above from -- `ubus call poe $whatnow`?

2) If I get another of those BCM PSEs dropped in my lap, what does this let me do -- map the ports more easily back to the case LEDs and their printed indications?

grobian:
So, this was just my take on things.  I think it's more debugging functionality, and it could be rewired to the debug function.  The only thing it does, is read the states as they are available.  It doesn't modify.  That part is what I attempted to do in my fork, but to no avail sofar.  I'll test on a GS1900-8HP once I can to see if changing the LED config (the final block) really works.  If so, config syntax can be introduced for this, when considered interesting enough.  With the HPE 1920s I've reached a dead end regarding trying to get PoE LEDs working (limit to my capabilities).

grobian:
> 1. What's the output above from -- `ubus call poe $whatnow`?

I don't understand this question.  I can say that `[ $whatnow == "leds" ]` and the result is the JSON dump you see here.

Hurricos:
> > 1. What's the output above from -- `ubus call poe $whatnow`?
> 
> I don't understand this question. I can say that `[ $whatnow == "leds" ]` and the result is the JSON dump you see here.

That was it -- thanks.

You could move it to e.g. `debug_leds` just for semantic clarity. I agree with the use-case, I've broken teeth on mapping crap and this is a fine use-case (I make no claims on the changeset itself quite yet).

grobian:
renamed to `debug_leds`

mrnuke (commented):
Skipping over the style issues.
The existing pattern is to:
- interpret the data in the reply handler
- report the value in the ubus handler

I would stick to the existing pattern and do any interpretation in the reply handler.

I am intrigued by the "error" string in the output for the HPE 1920-24G PoE. The LED replies do not have error info. What happened there?

grobian:
Can you give an example of interpretation you would like to see done in the reply handler?

The "error" string is because the controller actually returned error (01).  So the debug output is just showing exactly why the map table is empty.

mrnuke:
> Can you give an example of interpretation you would like to see done in the reply handler?

It would be parsed to a const string as the data is incoming:
```	
	led_cfg->sys_ok = GET_STR(reply[0x2] & 0x1, led_modes);
	led_cfg->in_gb = GET_STR(reply[0x3] & 0x3, led_modes);
```


> The "error" string is because the controller actually returned error (01). So the debug output is just showing exactly why the map table is empty.

That's not what the code does. It prints "error" if the first mapping doesn't start at zero. I don't think that's an error, not according to [[1]](https://svanheule.net/switches/software/broadcom_poe_control_protocol#get_port_led_map) .
```
	if (state->ledmaps[0].offset != 0)
		blobmsg_add_string(b, tmp, "error");
```

[1] https://svanheule.net/switches/software/broadcom_poe_control_protocol#get_port_led_map

grobian:
it is, because the offset should be a multiple of 8, so if 0 isn't 0, it means it is wrong

mrnuke:
I'm not sure I fully agree. Logic works for mapping of (0, 8, 16). It fails for (16, 8, 0), or (8, 0, 16), etc. If I'm trying to debug, and I see "error", I don't know what value caused the error. I then need to look at the actual packet, which I think would defeat the purpose of the "debug" query. I think the best result is to report the actual value in the `offset` field.

```
               "map_0_offset":  1,
                "lan8": 0,
                "lan7": 0,
```

You may also choose to not report the mapping at all, if it's disabled.

grobian:
> Skipping over the style issues. The existing pattern is to:
> 
>     * interpret the data in the reply handler
> 
>     * report the value in the ubus handler

I'm intrigued.  Why would you want to compute values every time you refresh the value (over and over and over again) as opposed to only do that work once you actually have someone/something that has an interest for that value?  And not to mention when you want to generate multiple views over the data, think like reporting temp in C or F, or a status/info view, which is really "user" friendly, but another view which is for debugging, and returns the "raw" values, or something?  Would you be of the opinion that one has to prepare for all of that during the reply?

mrnuke:
> I'm intrigued. Why would you want to compute values every time you refresh the value (over and over and over again)

It's a matter of design priorities. Pointer offsets and bitwise operations are a few clock cycles light. On the one hand, they can be deferred. On the other hand, we have readability by using consistent code patterns. Since the former seems like a micro-optimization, I choose the latter in this case.

Food for thought: the reply handler call rate is limited by the serial connection, while the rate of ubus calls is theoretically unbounded.

> you want to generate multiple views over the data, think like reporting temp in C or F

My comment on #44 was about naming the variable `temp_celsius`, or `pse_temp_celsius` to make the units and/or point of measurement clear. The purpose is not to export it in multiple units.


For the debugging info, I don't usually mind if it's curated or raw. My main concern is that there is a 1:1 mapping between the curated and raw value. For example a raw value of (0, or 1) can be curated to ("okay", "error").But when (1, 2, 3) are all mapped to "error", and all I get is "error", I don't know what caused it. So I have to go look at the packet dumps

grobian:
> Food for thought: the reply handler call rate is limited by the serial connection, while the rate of ubus calls is theoretically unbounded.

Ok -- the serial connection is already swamped at times, while calling ubus more often than the refresh rate the serial connection can produce is completely useless (you won't get any "fresher" data).  However, you can clearly correlate cpu usage versus call-rate.  The more you call, the higher the usage.

grobian:
I changed it such that the map shows offset error, and no mappings, or the offset and the retrieved mappings

Neustradamus:
Dear all,

What is the situation of this PR?

grobian:
I think this should be rejected/closed.  It was a nice experiment to extract settings info, but setting it doesn't work (AFAICT), and thus my impression is the infra is mostly ignored.  Either a PoE LED exists and is activated automatically similar to activity and link speed, or there exists a separate LED, that I've yet have to find a way to address/activate, but controlled from userspace.

Hurricos:
Mmmh ... reflexively replying to @Neustradamus, same vibe as [here](https://github.com/Hurricos/realtek-poe/pull/44#issuecomment-3476558243).

I get @grobian's initial use-case for it. I like taking people at their word (RE: "we should close this PR"), particularly when they've had time to tinker, think and have their feelings settle. I doubt @mrnuke feels strongly, but if he does, we can reopen.

mrnuke:
I have no issues with putting something like this in a debug packet. It would have to be rebased on top of the bcm/rtl dialect split. I'm not sure if such info is available outside the broadcom dialect.

PR #52, opened 2025-01-25
Use constant for frame size

svanheule:
Instead of hardcoding the frame size of 12 bytes everywhere in the code, use a constant to ensure it is used consistently. This fixes an issue with the "sendframe" command, which would only accept 9 bytes instead, resulting in the last two bytes of the frame being ignored.

Additionally, as it is frequently used, also define a constant for the checksum offset.

Fixes: 661ccb9a8629 ("realtek-poe: Add a ubus call to send frames")


mrnuke (approved):
! prefer a `CMD_SIZE - 1` instead of `OFFSET_CHECKSUM`., but either solution works.

PR #61, opened 2025-09-10
realtek-poe: remove commandline argument for baudrate

### LINKED ISSUE #60: commandline baudrate parameter does not work
The getopt parameter in main.c line 1279 expects either `d` or `s`, but the case statement in line 1285 expects `f` instead of `s`.
https://github.com/Hurricos/realtek-poe/blob/realtek-poe/src/main.c#L1279

Due to that, the -f switch is ineffective.

mrnuke:
Commit e58055c ("tek-poe: allow changing baudrate dialect in the config") introduced two ways to override the default baudrate. Having a commandline parameter to change the baudrate was useful during development, but should not have been committed otherwise. This function does not work due to a typo in the getopt() arguments.

The design philosophy of realtek-poe is that it should work with minimal configuration. Because an auto-detection mechanism is not currently implemented, the baudrate needs to be configured by the user. However doing so from two inputs could be confusing.

Remove the commandline argument to prevent confusion. The `force_baudrate` config option be used instead.

Fixes: e58055c ("tek-poe: allow changing baudrate dialect in the config")

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1276,14 +1276,12 @@ int main(int argc, char **argv)
 	ulog_open(ULOG_STDIO | ULOG_SYSLOG, LOG_DAEMON, "realtek-poe");
 	ulog_threshold(LOG_INFO);
 
-	while ((ch = getopt(argc, argv, "ds")) != -1) {
+	while ((ch = getopt(argc, argv, "d")) != -1) {
 		switch (ch) {
 		case 'd':
 			ulog_threshold(LOG_DEBUG);
 			poe.hardcore_hacking_mode_en = 1;
 			break;
-		case 'f':
-			baudrate = 115200;
 		}
 	}
 
@@ -1298,10 +1296,8 @@ int main(int argc, char **argv)
 	/* Users of poe_dialect assume the reverse mapping is computed. */
 	dialect_reverse_map(&poe.mcu.dialect);
 
-	/* Prefer '-s' argument over any config file option */
-	if (baudrate) {
-		/* Keep existing baudrate */
-	} else if (poe.config.forced_baudrate) {
+	/* Temporarily allow config file to override baudrate */
+	if (poe.config.forced_baudrate) {
 		baudrate = poe.config.forced_baudrate;
 	} else {
 		baudrate = 19200;


Neustradamus:
@Hurricos: Have you seen this @mrnuke PR?

Hurricos:
@BrainSlayer I know you use realtek-poe in DD-WRT (I haven't mentioned it yet, but I am disgusted that you do so).

Does losing the ... `-s`? ... option ... break your use-case?

Also, I've commented on the review, I don't think `-s` ever worked, as the `getopt` invoked `"ds"`, not `"df:"`.

src/main.c
```diff
@@ -1276,14 +1276,12 @@ int main(int argc, char **argv)
...
 	ulog_threshold(LOG_INFO);
 
-	while ((ch = getopt(argc, argv, "ds")) != -1) {
+	while ((ch = getopt(argc, argv, "d")) != -1) {
 		switch (ch) {
 		case 'd':
 			ulog_threshold(LOG_DEBUG);
 			poe.hardcore_hacking_mode_en = 1;
 			break;
-		case 'f':
```
Hurricos, line 1285: Wait ... `case 'f'`??

Did `-s` ever even work?

Also, browsing [`getopt(3)`](https://man7.org/linux/man-pages/man3/getopt.3.html), I'm coming up with a very "silly man question" -- wouldn't it have had to be `"df:"`, since `-f` takes (took) an argument?

I doubt @BrainSlayer ever used this option in DD-WRT, but [this is the only use-case I can think of for -f](https://github.com/Hurricos/realtek-poe/pull/61#issuecomment-3476568675), as DD-WRT doesn't use uci. (Whether we care about non-OpenWrt uses is up-for-debate, but I think we might if they exist.)

mrnuke: The `case 'f':` never worked.

Hurricos: LOL. Merged.

BrainSlayer:
> @BrainSlayer I know you use realtek-poe in DD-WRT (I haven't mentioned it yet, but I am disgusted that you do so).
> 
> Does losing the ... `-s`? ... option ... break your use-case?
> 
> Also, I've commented on the review, I don't think `-s` ever worked, as the `getopt` invoked `"ds"`, not `"df:"`.

no. i dont use the commandline. (only for debugging). i removed the uci config parser  and replaced it with a own config reader. i mean for quick trying and debugging its usefull. but i rather would like a parameter with arguments. so you can play with different baudrates in a quick way instead of a forcefull 115200 baudrate. so keeping a argument like that makes totally sense for developers.

PR #65, opened 2025-11-09
build: require CMake >= 3.10 due to dropped legacy support

hauke:
CMake version 4.0 and later require minimum version of 3.5 or later. Update to minimum version 3.10 which is the last not deprecated minimum version.

CMake 3.10 was released in November 2017 and is included in Ubuntu 18.04.


mrnuke (approved):
What a missed opportunity to make a math reference and require 3.14

